PR #12, opened 2026-05-18
text_field: secure: true for masked password input

C-Sinclair:
## Summary

Adds a `secure: true` prop to `text_field`. iOS renders the field as a SwiftUI `SecureField` (masked input) instead of `TextField`. Cleartext still reaches the BEAM via `on_change` — apps hash/store the value as normal.

```elixir
%{type: :text_field, props: %{
  value: assigns.password,
  placeholder: "Password",
  secure: true,
  on_change: {self(), :password_changed}
}}
```

The prop flows through `Mob.Renderer.prepare_props/4`'s generic pass-through; no renderer changes were needed. Native side: `MobNode.isSecure` BOOL, parsed in `mob_nif.m`, branched on in `MobRootView.swift` via a `@ViewBuilder` field switch.

### Version / CHANGELOG

CHANGELOG entry under `## [Unreleased]`; no `mix.exs` bump. Matches the repo's pattern of feature-commits-without-bumps and a separate ride-along `Bump to X.Y.Z` commit at release-cut time (e.g. b974398).

### Android

This PR is iOS-only. Mob's library doesn't own Android Compose UI — that lives per-app, generated from `mob_new`'s `MobBridge.kt.eex` template. A companion PR against [GenericJam/mob_new](https://github.com/GenericJam/mob_new) adds `PasswordVisualTransformation` + `KeyboardType.Password` to the template.

Until that ships, `secure: true` is a graceful no-op on Android: the prop travels through the renderer but the Compose `TextField` ignores it. No breakage for unpatched apps.

### Why no reveal-toggle ("eye" button) yet

Worth its own PR. The iOS implementation needs `ZStack`-and-opacity to retain focus across `SecureField`/`TextField` swaps — the current `@ViewBuilder if/else` (and any naive toggle) tears the view down on every flip, dropping focus and dismissing the keyboard. Pairs with a `trailingIcon` slot in Compose. Out of scope here.

## Test plan

- [x] `mix test` — 794 passed (27 doctests, 767 tests), 43 excluded. Two new renderer tests cover `secure: true` serialising and absence-by-default.
- [x] `mix format --check-formatted` — clean
- [x] `mix credo --strict` — clean (949 mods/funs, 0 issues)
- [x] `mix compile --warnings-as-errors` — clean
- [x] `mix erlfmt --check src/` — clean
- [x] `xcrun clang-format --dry-run -Werror ios/*.m ios/*.h` — clean
- [x] Verified on iOS sim (iPhone 17, iOS 26.2): `SecureField` renders dots; BEAM `:password` assign holds cleartext (`"hunter2"`); `Length: 7` label updates live.
- [x] Verified Android contract end-to-end with the matching `MobBridge.kt` patch applied locally: dots render, cleartext reaches BEAM, length label updates. (Template patch lands in the mob_new companion PR.)
- [ ] `swiftlint ios/` — not installed locally; the Swift diff is mechanical (single expression swapped for an `@ViewBuilder` computed property).

## Files

- `ios/MobNode.h` — `@property(nonatomic, assign) BOOL isSecure`
- `ios/mob_nif.m` — parses `props[@"secure"]` alongside `keyboard` / `return_key`
- `ios/MobRootView.swift` — `@ViewBuilder var field` picks `SecureField` or `TextField`; existing modifier chain applies uniformly to both
- `test/mob/renderer_test.exs` — pass-through and default-absent tests
- `CHANGELOG.md` — entry under `## [Unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
 
 ---
 
+## [Unreleased]
+
+### Added
+- `text_field` now accepts a `secure: true` prop. iOS renders the field
+  as a SwiftUI `SecureField` (masked input) instead of the plain
+  `TextField`. The prop flows through the existing renderer
+  passthrough; cleartext still reaches the BEAM via `on_change` so apps
+  can hash/store the value as normal. Android consumes the same prop
+  via `PasswordVisualTransformation` once `mob_new`'s `MobBridge.kt.eex`
+  template is updated in a companion PR — until then the prop is a
+  graceful no-op on Android (renders as a regular field), no breakage.
+
+  Reveal-toggle ("eye" button) is intentionally deferred — its
+  interaction with SwiftUI focus retention requires a `ZStack`-and-opacity
+  rebuild of `MobTextField` and warrants its own change.
+
 ## [0.6.7]
 
 ### Added
--- a/ios/MobNode.h
+++ b/ios/MobNode.h
@@ -125,6 +125,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, copy, nonnull)
     NSString *keyboardTypeStr; // "default","number","decimal","email","phone","url"
 @property(nonatomic, copy, nonnull) NSString *returnKeyStr; // "done","next","go","search","send"
+@property(nonatomic, assign) BOOL isSecure;                 // mask input (SecureField on iOS)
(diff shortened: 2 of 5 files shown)


GenericJam:
Thanks @C-Sinclair — landed as 290254b in mob 0.6.15 with your authorship preserved. Paired with mob_new#6 (Android Compose side) landed as 4a41f47 in mob_new 0.3.8, so `secure: true` now works on both platforms in the same release.

CHANGELOG entries merged into the new [0.6.15] block (and mob_new [0.3.8]). Closing since the commit is on master.

PR #14, opened 2026-05-18
iOS: file-only :inet_db lookup by default in Mob.App.start/0

C-Sinclair:
## Summary

`Mob.App.start/0` now switches `:inet_db`'s lookup chain to `[:file]` and seeds `localhost` on iOS, before any user code runs. The first `Node.connect` / `:erpc.call` / `gen_tcp.connect/3` with a binary host no longer crashes the calling process with `:badarg`.

## Why

BEAM's default `:native` hostname resolution spawns the `inet_gethost` port program. iOS's app sandbox forbids `execve` of any binary the app didn't get a special pass for, so the spawn fails categorically on device. The simulator hits the same failure for a related-but-distinct reason — `inet_gethost` doesn't live at the path BEAM expects under the mob iOS-sim OTP layout. Either way, the lookup dies and any caller through `:inet.getaddr/2` (distribution, RPC, TCP-by-name) gets `:badarg` and exits.

Every iOS mob app today has to do this dance manually in `on_start/0` before anything touches a hostname:

```elixir
:inet_db.set_lookup([:file])
:inet_db.add_host({127, 0, 0, 1}, [~c"localhost"])
```

Easy to forget, impossible to test for, and bites every new consumer the first time they call `:erpc.call`. The framework should own this.

## Composability with `Mob.DNS`

Pure addition. `Mob.DNS.configure_pure_beam/1` keeps working — it upgrades `[:file]` → `[:file, :dns]` and seeds fallback nameservers. The file-table entries this PR adds stay first in the chain and keep winning, so the two layers compose:

- **Framework default** — `[:file]`, distribution + local-loopback TCP work. No DNS.
- **App opts in** to `Mob.DNS.configure_pure_beam/1` for outbound HTTP — chain becomes `[:file, :dns]`, fallback nameservers (Google + Cloudflare by default) handle public hosts.
- **App opts in** per-host to `Mob.DNS.resolve/1` for VPN / mDNS / search-domain semantics — Apple resolver via NIF, result seeded in the file table, always wins because `:file` is first.

Apps that *want* the default OS resolver back (and were somehow getting away with it on iOS) can call `:inet_db.set_lookup([:native])` in their own `on_start/0`, which runs after `start/0`.

## Why `[:file]` not `[:file, :dns]` by default

Conservative — the irreducible fix is preventing the `inet_gethost` spawn. Adding `:dns` to the framework default would mean every iOS app pays for fallback-nameserver state it may not use, and apps that talk to an internal-only backend on a fixed IP get a slightly-noisier `inet_db`. `Mob.DNS.configure_pure_beam/1` is the right opt-in for apps that need DNS; the framework just guarantees the crash doesn't happen.

## Test plan

- [x] `mix test` — 798 passed (27 doctests, 771 tests), 43 excluded. New `test/mob/app_test.exs` verifies the host-BEAM (NIF-unloaded) path is a no-op and that the function is idempotent.
- [x] `mix format --check-formatted` — clean
- [x] `mix credo --strict` — clean (952 mods/funs, 0 issues)
- [x] `mix compile --warnings-as-errors` — clean
- [x] `mix erlfmt --check src/` — clean
- [ ] iOS sim end-to-end — needs a deployed app; can't reproduce in this PR's CI. Verified manually that without the fix the first `:erpc.call` from a sim app crashes with `inet_gethost` `open_port` `:badarg`, and with the fix the same call succeeds.

## Related

Companion to the other small contributions in this series — `text-field-secure` (#12), `fix/ios-column-fill-height`, `scroll-on-refresh`, `nav-transition-override`. Each stands alone; land in any order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- iOS: `Mob.App.start/0` now switches `:inet_db` to file-only lookup and seeds `localhost` before any user code runs — BEAM's default `:native` lookup tries to `execve` the `inet_gethost` port program, which the iOS sandbox refuses, crashing the first `Node.connect` / `:erpc.call` / `gen_tcp.connect/3` with `:badarg`. Apps no longer need to set the lookup chain themselves; `Mob.DNS.configure_pure_beam/1` still composes on top for outbound DNS. See `guides/dns_on_ios.md`.
+
 ## [0.6.7]
 
 ### Added
--- a/guides/dns_on_ios.md
+++ b/guides/dns_on_ios.md
@@ -34,6 +34,25 @@ first in the chain.
 
 ---
 
+## What mob already does for you
+
+Before any of this matters, `Mob.App.start/0` runs
+`Mob.App.configure_ios_inet_db/0` on iOS (both simulator and device),
+which switches the lookup chain to `[:file]` and seeds `localhost`.
+That's the minimum needed so `Node.connect`, `:erpc.call`, and
+`gen_tcp.connect/3` with a binary host don't crash the calling
+process on the first lookup. Apps don't have to do anything for
+distribution and local-loopback TCP to work.
+
+`Mob.DNS.configure_pure_beam/1` is the next step *on top of that* —
+it upgrades the chain to `[:file, :dns]` and seeds fallback
+nameservers so outbound HTTP to public hosts works. The framework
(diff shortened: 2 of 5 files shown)


GenericJam:
Thanks @C-Sinclair — landed as 11752a4 in mob 0.6.15 with your authorship preserved. CHANGELOG resolution merged both your `### Fixed` entry and the paired #12 `### Added` entry under the new [0.6.15] section.

The framework now owns the `:inet_db` file-only default so new iOS apps don't have to remember the two-line `on_start/0` dance. `Mob.DNS.configure_pure_beam/1` still composes on top for apps that want DNS. Closing since the commit is on master.

Issue #17, opened 2026-05-20
Template collocation


clsource:
Is possible to have a collocated mob file In the same spirit such as [LiveView](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#module-template-collocation) ?

Example.

Is the same file as the demo, but without the render function.

**home_screen.ex**

```elixir
defmodule MiApp.HomeScreen do
  @moduledoc "Landing screen for MiApp."
  use Mob.Screen

  def mount(_params, _session, socket) do
    theme = Mob.State.get(:theme, :obsidian)
    Mob.Theme.set(theme_to_module(theme))
    {:ok, Mob.Socket.assign(socket, :theme, theme)}
  end

  def handle_info({:tap, :open_text}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.TextScreen)}
  end

  def handle_info({:tap, :open_list}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.ListScreen)}
  end

  def handle_info({:tap, :open_dice}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.DiceScreen)}
  end

  def handle_info({:tap, :open_webview}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.WebViewScreen)}
  end

  def handle_info({:tap, :open_audio}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.AudioScreen)}
  end

  def handle_info({:tap, :open_camera}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.CameraScreen)}
  end

  def handle_info({:tap, :open_storage}, socket) do
    {:noreply, Mob.Socket.push_screen(socket, MiApp.StorageScreen)}
  end

  def handle_info({:tap, :theme_obsidian}, socket) do
    Mob.Theme.set(Mob.Theme.Obsidian)
    Mob.State.put(:theme, :obsidian)
    {:noreply, Mob.Socket.assign(socket, :theme, :obsidian)}
  end

  def handle_info({:tap, :theme_citrus}, socket) do
    Mob.Theme.set(Mob.Theme.Citrus)
    Mob.State.put(:theme, :citrus)
    {:noreply, Mob.Socket.assign(socket, :theme, :citrus)}
  end

  def handle_info({:tap, :theme_birch}, socket) do
    Mob.Theme.set(Mob.Theme.Birch)
[...]

PR #22, opened 2026-05-22
feat: support collocated screen templates

dl-alexandre:
## Summary
- compile a sibling `.mob.heex` file into `render/1` for screens that use `Mob.Screen`
- preserve explicit `render/1` definitions when a screen already provides one
- cover the generated render path with a collocation regression test

## Tests
- `ELIXIR_COMPILER_OPTIONS="--no-parallel" mix test test/mob/screen_collocation_test.exs test/mob/screen_test.exs`


clsource:
closes #17

PR #35, opened 2026-05-28
docs: activate mermaid dark theme for hexdocs

clsource:
Improved hexdocs rendering


clsource:
This PR solves two problems in Hexdocs

1. Does not render diagrams in architecture

imagen

This is due to the odd code used for triggering renders.
It was used a standard code for rendering all `.mermaid` code blocks.

2. Added dark theme
imagen

The current white theme does not look good in hexdocs dark mode.
Maybe it can be improved to be the same as the hexdocs option.

imagen

clsource:
maybe is better other approach

Issue #25, opened 2026-05-22
Init a barebones project with mob.new


clsource:
In the current `mob.new` command it will init with example code.
But I have to delete them everytime I would need a new project.

Adding an option such as `mob.new myproject --demo` will keep the current way with demo files.

<img width="225" height="242" alt="Image" src="https://github.com/user-attachments/assets/80a4eaf9-e6a3-4f3a-bfce-a5ee11e55d13" />

`mob.new myproject` would init a blank project with just the barebones needed
to compile an app.

- app.ex
- repo.ex
- home_screen.ex

Other idea is using `mob.new myproject --blank` to allow a barebones configuration, but I guess that 
is more common to use a blank template than a template filled with demo files.

clsource:
moved to mob_new repo

PR #54, opened 2026-06-29
Mob.Audio output probes: verify sound is actually working

GenericJam:
## What

The audio analog of `screenshot/3`: read-only probes that answer **"is sound actually coming out right now."**

- **`Mob.Audio.output_status/0`** → `%{volume, muted, route, other_audio}`. Cheap, no permission. Catches the common silence causes (muted, volume 0, dead route). iOS `AVAudioSession` / Android `AudioManager`.
- **`Mob.Audio.output_level/1`** → `{rms_db, peak_db}` | `:silent` | `{:error, reason}`. Reads actual signal energy — the part `output_status` and `adb` can't answer. `source: :mob` (default) meters **Mob.Audio's own player** (iOS `AVAudioPlayer` metering; Android `Visualizer` on the player's own session, needs runtime `RECORD_AUDIO`). `source: :mix` returns `{:error, :unsupported_on_platform}`.

## Scope correction from device verification

The original design had `:mix` tap the **global output mix** to catch audio from native players that bypass `Mob.Audio` (a game's own `AudioTrack`). **Device testing on a moto g power (2021), Android 11, disproved this:** a session-0 `Visualizer` fails with `ERROR_NO_INIT` even with `RECORD_AUDIO` + `MODIFY_AUDIO_SETTINGS` — global output capture is privileged (signature `CAPTURE_AUDIO_OUTPUT`, or MediaProjection `AudioPlaybackCapture`). iOS forbids it by sandbox.

So in-app these probes verify **your own** audio. True global/foreign-app capture is deferred to a separate **MediaProjection-based capture plugin** intended as a test-environment dependency (Android-only). For "is a bundled game's audio working" without that plugin, `adb shell dumpsys media.audio_flinger` (active track + underruns) is the answer.

## Device verification (moto g power 2021, Android 11) ✅

dist-RPC into a real build:
- App **boots** with the new NIF table (stable pid) — boot-critical for a native-table mismatch.
- `output_status/0` → `%{route: :speaker, volume: 0.2, muted: false, other_audio: false}`.
- `output_level(:mob)` while a looped tone played → `{-34.8, -31.8}`; idle / after stop → `{:error, :not_playing}`.
- `output_level(:mix)` → `{:error, :unsupported_on_platform}`.
- Without `RECORD_AUDIO` granted → `{:error, :needs_record_audio}`.

## How

Native wiring mirrors `screenshot/3` + `open_settings/1`: `-export`/`-nifs`/stub in `mob_nif.erl`, native table entries in `mob_nif.zig` + `mob_nif.m`, `cacheOptional` + null-guard for the app-owned Android bridge methods (drift-safe boot). `output_level` is a dirty IO NIF. Android returns a length-coded `float[]` (len 2 = level; len 1 = error code) the NIF maps to atoms; decoding (route codes, `:silent`/`:error`) is pure Elixir (`decode_status/1`, `decode_level/1`), unit-tested on host. Companion Android Kotlin: GenericJam/mob_new#25.

Host: mob test suite green, formatters clean (mix format, erlfmt, clang-format, zig fmt). Decision record + follow-up plugin scope: `decisions/2026-06-29-audio-output-probes.md`.

### DIFF
--- a/android/jni/mob_nif.zig
+++ b/android/jni/mob_nif.zig
@@ -2139,10 +2139,14 @@ export fn nif_audio_output_status(
 }
 
 // nif_audio_output_level/1 — {RmsDb, PeakDb} as two doubles, or an error atom.
-// Source is "mix" (global output mix, needs RECORD_AUDIO) | "mob" (Mob's own
-// player). MobBridge.audioOutputLevel(String) returns float[2] = [rms_db,
-// peak_db] on success, or null on failure (no permission / no signal /
-// unsupported), which we surface as an atom Mob.Audio maps to {:error, _}.
+// Source is "mob" (Mob's own player session) | "mix". The global output mix is
+// privileged on modern Android (a normal app gets ERROR_NO_INIT attaching a
+// Visualizer to session 0), so "mix" is unsupported here — global device-audio
+// capture lives in a separate MediaProjection-based plugin. MobBridge returns:
+//   float[2] = [rms_db, peak_db]  → success
+//   float[1] = [code]             → 1 unsupported_on_platform, 2 needs_record_audio,
+//                                    3 not_playing (no active Mob.Audio player)
+//   null                          → generic error
 export fn nif_audio_output_level(
     env: ?*erts.ErlNifEnv,
     argc: c_int,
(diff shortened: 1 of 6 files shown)


GenericJam:
Companion Kotlin bridge methods: GenericJam/mob_new#25. Merge together.

GenericJam:
## Review — approve with follow-ups (FFI seam clean, drift-safe)

Traced both new NIFs (`audio_output_status/0` `()[F`, `audio_output_level/1` `(Ljava/lang/String;)[F`) across every layer — Elixir → `mob_nif.erl` → zig table/descriptor → Kotlin `@JvmStatic` → iOS nif table. Arities/descriptors match, dirty flags match. And unlike the input-metering PR #67, this uses **`cacheOptional` + null-guards**, so a drifted app bridge degrades gracefully (no boot crash) — nice. `decode_status`/`decode_level` host tests are meaningful.

Two semantic gaps worth tracking (both verified in the diff, not blockers):

**1. `:silent` never fires on Android.** `decode_level({_rms, peak}) when peak <= -120.0 -> :silent`, but this PR's own doc notes the Android `Visualizer` floor is **-96 dB**. -96 > -120, so pushed silence on Android decodes as a real signal — `:silent` is effectively iOS-only. A caller using `output_level(...) == :silent` to detect silence gets a false negative on Android. Suggest raising the threshold above the Android floor (e.g. `-95.0`, or normalize per-platform) and adding a `{-96.0, -96.0}` host test (the suite tests `{-160,-160}`/`{-130,-120}` but never the Android floor).

**2. `output_level(:mob)` doesn't see `play_at`.** iOS meters `g_audio_player` (the `play/1` player); the sample-accurate `play_at` path uses separate scheduled players that never get `meteringEnabled`. Android is analogous (`audioPlayer` MediaPlayer vs the `AudioTrack` scheduling path). So after `Mob.Audio.play_at`, `output_level(:mob)` reports `:not_playing`/`:not_metering` while sound is audibly playing — notably the feature's motivation (an app driving its own AudioTrack) is exactly the path the probe can't see. Suggest metering scheduled players too, or documenting the `play/1`-only scope.

Minor: the title "verify sound is actually working" is a bit stronger than what two probes measure (signal-present + config-sane, not emission-at-the-speaker) — the module docs are honest about this, so just don't let the title over-set expectations.

Merge with mob_new #25 (Android needs its bridge methods). CI green, base current.

_(FFI parity, `cacheOptional`, the -120/-96 gap, and the `g_audio_player`/`play_at` split are confirmed against PR-head; iOS metering values are not device-verified per the PR.)_

PR #56, opened 2026-07-01
MOB-5: ~MOB @foo raises a clear error when assigns is not in scope

GenericJam:
Fixes **MOB-5**. Follow-up to the 0.7.11 LiveView-ergonomics release.

## Problem

0.7.11 added `@foo` → `assigns.foo` sugar in `~MOB`. But the idiomatic way to factor mob UI — including the guide's own example — is **plain functions with positional args**:

```elixir
def card(title, children), do: ~MOB"<Text text={title} />"
```

There's no `assigns` in such a function. The moment a user follows the new "@assigns shorthand" docs and writes `@title` inside one, it expands to `assigns.title` → cryptic **"undefined variable assigns"**. (Same trap if a `render` param is named `socket` instead of `assigns`.) A user hit exactly this: a wave of "no assigns" errors, then "unused assigns" warnings while trying to bridge the two styles.

## What Phoenix does

`Phoenix.Component.sigil_H` guards with `Macro.Env.has_var?(__CALLER__, {:assigns, nil})` and raises *"~H requires a variable named assigns..."* — a clear compile error instead of a raw undefined-variable. mob's sigil has `__CALLER__` too, so it can do the same.

## Fix

- **`Mob.Sigil`:** before rewriting `@foo`, check `Macro.Env.has_var?(caller, {:assigns, nil})`. If `assigns` isn't bound, raise a `CompileError` naming the fix — *"`@title` requires a variable named `assigns` in scope … use `{title}` instead of `@title`."* Applies to attrs, `{expr}` children, and `:if`/`:for`. Only `@`-using templates trigger it; a static `~MOB(<Text text="hi"/>)` in a positional-arg helper still compiles (unlike `~H`, which always needs assigns).
- **Docs:** scope `@` to `render(assigns)`, steer composites to positional `{var}`, and add the common-mistake note (including "name the render param `assigns`").

## Tests
`@foo` with no assigns raises (asserts the message names `{title}`); `:if={@flag}` with no assigns raises; a static template compiles without assigns; a positional-arg helper (`{title}`) compiles + runs without assigns. Existing `@assign` tests (which bind `assigns` locally) stay green. Full suite: 964 pass; format + credo clean.

## Not included
No version bump — release timing is the maintainer's call (would be 0.7.12).

### DIFF
--- a/guides/components.md
+++ b/guides/components.md
@@ -45,7 +45,7 @@ The sigil borrows three authoring idioms from Phoenix HEEx, so screens read the 
 
 ### `@assigns` shorthand
 
-Inside any `{...}` expression, `@foo` rewrites to `assigns.foo` (this happens at compile time). It works in attribute values, `{expr}` children, and the `:if`/`:for` directives below. Nested access like `@user.name` works too.
+Inside a `{...}` expression, `@foo` rewrites to `assigns.foo` at compile time. It works in attribute values, `{expr}` children, and the `:if`/`:for` directives below. Nested access like `@user.name` works too.
 
 ```elixir
 def render(assigns) do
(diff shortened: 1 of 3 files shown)


GenericJam (commented):
Reviewed the diff, tests, and CI. **Recommend merge.**

**Correctness:** the guard sits inside the `prewalk`, so it only fires when a `@name` node is actually present — static templates and positional-arg helpers using `{var}` compile untouched. `caller` is threaded correctly from the sigil `__CALLER__`, and `Macro.Env.has_var?(caller, {:assigns, nil})` is the same check `Phoenix.Component.sigil_H` uses, so it is proven.

**Tests:** targeted and real — `@foo` without assigns raises (asserting the message names `{title}`), `:if={@flag}` raises, a static template compiles, and a positional helper both compiles and runs. Guard + happy path both covered.

**CI:** fully green (Elixir 1.19/OTP 28, native formatters, mix_audit, GitGuardian); mergeable/CLEAN.

**Minor, non-blocking:** the guard also fires for a genuine module attribute `@title` used inside a sigil in a non-`assigns` function. That is pre-existing `@`-shadowing (same as HEEx) and the new error is still clearer than the old undefined-variable crash, so not worth holding the PR.

**Coordination note (docs anchor):** the new blockquote links to `[function composites](#pure-elixir-composite-components)`. PR #57 renames that section heading to "Defining your own components", which removes that anchor. I am updating #57 to keep the `#pure-elixir-composite-components` anchor alive so this link does not dangle regardless of merge order. Suggest merging this (#56) first, then #57 rebases on top.

(Posting as a comment rather than an approval — GitHub blocks approving your own PR.)

Issue #28, opened 2026-05-24
iOS build error: invalid option: -Dproject_swift_sources


clsource:
MacOS: Tahoe 26.3
Xcode: 26.5
Mob: 0.6.18

Using a new project it cannot build.

```bash
$ mix mob.new calculator
$ cd calculator
$ mix mob.install
$ xcrun simctl boot "iPhone 16e"
$ open -a Simulator
```

`$ mix mob.doctor`

```text
Tools
  ✓ version manager — mise (/Users/camilo/.local/bin/mise)
  ✓ Elixir — 1.20.0-rc.5
  ✓ OTP — 29 (ERTS 17.0)
  ✓ Hex — 2.4.2
  ✓ epmd — /Users/camilo/.local/share/mise/installs/erlang/29.0/erts-17.0/bin/epmd
  ✓ adb — /Users/camilo/Library/Android/sdk/platform-tools/adb
  ✓ xcrun — Xcode 26.5
  ✓ zig — 0.16.0
  ✓ java — openjdk version "21.0.11" 2026-04-21 LTS
  ✓ Android SDK — /Users/camilo/Library/Android/sdk
  ✓ Android NDK — 27.2.12479018 (recommended) ✓
  ✓ python3 — /opt/homebrew/bin/python3
  ✓ rsync — /usr/bin/rsync
  ✓ ideviceinfo — /opt/homebrew/bin/ideviceinfo
```

`$ mix mob.deploy --native --ios`

```text
$ mix mob.deploy --native --ios

Fetching dependencies...
Resolving Hex dependencies...
Resolution completed in 0.048s
Unchanged:
  bandit 1.11.1
  bunt 1.0.0
  cc_precompiler 0.1.11
  credo 1.7.18
  db_connection 2.10.1
  decimal 3.1.0
  ecto 3.14.0
  ecto_sql 3.14.0
  ecto_sqlite3 0.24.0
  elixir_make 0.9.0
  eqrcode 0.2.1
  ex_ast 0.12.0
  ex_slop 0.4.1
  exqlite 0.36.0
  file_system 1.1.1
  finch 0.22.0
  glob_ex 0.1.11
  hpax 1.0.3
  igniter 0.8.0
  jason 1.4.5
  mime 2.0.7
  mint 1.8.0
  mix_audit 2.1.5
  mob 0.6.18
  mob_dev 0.5.11
  nimble_options 1.1.1
  nimble_parsec 1.4.2
  nimble_pool 1.1.0
[...]

PR #62, opened 2026-07-05
MOB-14: network / connectivity state in Mob.Device (query + events)

GenericJam:
Adds a cross-platform **network / connectivity state** capability to `Mob.Device`, alongside `battery_state`/`thermal_state`. Part of **MOB-14** (cross-repo; companion **[mob_new #28](https://github.com/GenericJam/mob_new/pull/28)** adds the Android `NetworkCallback`).

## API
```elixir
Mob.Device.network_state()
# iOS     => %{online: true, transport: :wifi, expensive: false, validated: :unavailable, constrained: false}
# Android => %{online: true, transport: :wifi, expensive: false, validated: true, constrained: :unavailable}
Mob.Device.online?()          #=> true
# transport: :wifi | :cellular | :wired | :other | :none
# expensive: metered link (cellular / hotspot) — defer big transfers
# validated: Android real-internet probe confirmed (false on captive portal); :unavailable on iOS
# constrained: iOS Low Data Mode on; :unavailable on Android
# — each single-platform signal returns :unavailable where the OS can't answer, never a misleading false

Mob.Device.subscribe([:network])
# => {:mob_device, :connectivity_changed, %{online: _, transport: _, expensive: _}}
```

## Wiring (mirrors the battery/thermal + orientation-event seam)
| Layer | Change |
|--|--|
| `lib/mob/device.ex` | `network_state/0`, `online?/0`, `:network` category, `connectivity_changed` event + `category_for/1` |
| `src/mob_nif.erl` | `device_network_state/0` in `-export` / `-nifs` / stub |
| `ios/mob_nif.m` | process-lifetime `NWPathMonitor` caches the snapshot + pushes `connectivity_changed`; query reads the cache. `Network.framework` autolinks via `-fmodules`; no permission/plist key |
| `android/jni/mob_nif.zig` + `mob_beam.h` | int-code-cached query NIF + `mob_send_connectivity_changed` export for the `beam_jni.c` trampoline |

## Verification — both platforms device-verified
- ✅ **iOS** (booted simulator): `network_state/0` → `%{online: true, transport: :wifi, expensive: false}` over dist; `:network` in `categories/0`. Native ObjC compiled+linked via real `mix mob.deploy --native`.
- ✅ **Android** (moto g power 2021): after a native build (compiled the Zig + `beam_jni.c` + Kotlin, `libnetcheck2.so` for all ABIs), `network_state/0` → `%{online: true, transport: :wifi, expensive: false}` over dist. Device verification caught a dangling-pointer bug (cached JNI string released by the trampoline → `transport: :"EMU=beam"`); fixed by caching an int transport code (last commit).
- ✅ Full suite **974** green incl. `nif_declaration_test` cross-layer guard; clang-format / zig fmt / erlfmt / credo --strict clean.

> CI does not compile native code — both native builds were verified locally on real targets.

### DIFF
--- a/android/jni/mob_beam.h
+++ b/android/jni/mob_beam.h
@@ -151,9 +151,9 @@ void mob_send_color_scheme_changed(const char *scheme);
 // Deliver {:mob_device, :connectivity_changed, %{online, transport, expensive}}
 // to the dispatcher pid registered via Mob.Device. Called from beam_jni.c's
 // nativeNotifyConnectivity when the ConnectivityManager.NetworkCallback fires.
-// `online`/`expensive` are 0/1; `transport` is
+// `online`/`expensive`/`validated` are 0/1; `transport` is
 // "wifi" | "cellular" | "wired" | "other" | "none".
-void mob_send_connectivity_changed(int online, const char *transport, int expensive);
+void mob_send_connectivity_changed(int online, const char *transport, int expensive, int validated);
 
 // mob_beam.h additions for Mob.Bt
 //
--- a/android/jni/mob_beam.h
+++ b/android/jni/mob_beam.h
@@ -148,6 +148,13 @@ void mob_send_component_event(int handle, const char *event, const char *payload
 // `scheme` must be "light" or "dark".
 void mob_send_color_scheme_changed(const char *scheme);
 
+// Deliver {:mob_device, :connectivity_changed, %{online, transport, expensive}}
+// to the dispatcher pid registered via Mob.Device. Called from beam_jni.c's
+// nativeNotifyConnectivity when the ConnectivityManager.NetworkCallback fires.
+// `online`/`expensive` are 0/1; `transport` is
+// "wifi" | "cellular" | "wired" | "other" | "none".
+void mob_send_connectivity_changed(int online, const char *transport, int expensive);
+
 // mob_beam.h additions for Mob.Bt
 //
 // Append these to the existing mob_beam.h, after the
--- a/android/jni/mob_beam.h
+++ b/android/jni/mob_beam.h
@@ -148,7 +148,8 @@ void mob_send_component_event(int handle, const char *event, const char *payload
 // `scheme` must be "light" or "dark".
 void mob_send_color_scheme_changed(const char *scheme);
 
-// Deliver {:mob_device, :connectivity_changed, %{online, transport, expensive}}
+// Deliver {:mob_device, :connectivity_changed,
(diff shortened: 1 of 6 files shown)


GenericJam:
Reviewed the code (not just the PR text) — **LGTM, endorse with two non-blocking nits.**

Verified: the FFI chain is fully consistent (Kotlin `nativeNotifyConnectivity(Boolean,String,Boolean,Boolean)` → beam_jni.c `(jboolean,jstring,jboolean,jboolean)` → zig `mob_send_connectivity_changed(c_int, ?[*:0]const u8, c_int, c_int)` → `mob_beam.h` prototype, all in order); the int-transport-code cache correctly fixes the dangling-JNI-string bug; iOS `mob_make_network_map` is correctly aligned (5 keys, `validated → :unavailable`); and the `:unavailable`-not-false design + consistent 5-key shape across platforms is a great call. Elixir seam has real coverage in device_test.exs.

Nits (neither blocks merge):

1. **Android/iOS atomicity parity.** iOS caches with `_Atomic(...)` + `atomic_load/store`, but the Android zig globals (`var g_net_online: bool`, `g_net_transport`, …) are plain — written by the JNI trampoline thread, read by the scheduler thread in `nif_device_network_state`. Real risk is near-zero (word-aligned reads don't tear on arm64; connectivity is eventually-consistent), but for parity + strict correctness consider `std.atomic.Value(...)`.

2. **Event-payload docs understate it.** The `mob_beam.h` comment, PR body, and the `subscribe/1` docstring in device.ex all say the event is `%{online, transport, expensive}` (3 keys), but both platforms actually deliver the full 5-key map (`networkStateMap` / `mob_make_network_map`). Extra keys are harmless (arguably better), but worth correcting the docstring so subscribers know `validated`/`constrained` are there too.

GenericJam:
Both nits addressed:

1. **Atomicity parity** — Android globals converted to `std.atomic.Value(bool)`/`Value(c_int)` with `.load/.store(.monotonic)`, matching the iOS `_Atomic` half. (The other Android device globals — `g_last_orientation` etc. — stay plain; that's pre-existing and out of scope here.)
2. **Docs** — device.ex is already 5-key in the `network_state/0` doc; clarified the `:connectivity_changed` line to say `state` is the same full map. Fixed the stale 3-key `mob_beam.h` comment to `%{online, transport, expensive, validated, constrained}`.

Rebuilding Android to recompile the atomic change + re-verifying `network_state/0` on the moto g before merge.

PR #66, opened 2026-07-05
MOB-20: keep-awake / idle-timer support in Mob.Device

GenericJam:
## What

Adds `Mob.Device.keep_awake/1` — prevent the screen auto-dimming/locking while enabled. This is the `mob` half; the Kotlin `MobBridge.keepAwake` method is the paired **GenericJam/mob_new** PR.

Linear: **MOB-20** (Tier-3 gap from the capability audit; high value-per-effort).

```elixir
Mob.Device.keep_awake(true)   # screen stays lit
Mob.Device.keep_awake(false)  # release
```

Today mob could only *observe* screen on/off, not prevent sleep — this closes that for video / reader / navigation screens.

## Design (see `decisions/2026-07-04-keep-awake.md`)

- **Core `Mob.Device`, not a plugin** — permission-free, universal device-state toggle, exactly like `lock_orientation/1`. Matches the `Mob.Device` convention (`keep_awake(on?) :: :ok`), not the socket-passthrough style of `Mob.Torch`/`Mob.Haptic`.
- **Write-only, re-assert on resume** — no getter (it's OS state); the flag is app-scoped and cleared on background, documented in the moduledoc.
- **Graceful drift** — the Android bridge method uses `cacheOptional` + null-guard (mirrors `orientationLock`), so an app whose generated `MobBridge.kt` predates it simply no-ops rather than failing to load.

## Layers

- `lib/mob/device.ex` — `keep_awake/1`.
- `src/mob_nif.erl` — `device_keep_awake/1` NIF decl + stub.
- `ios/mob_nif.m` — `nif_device_keep_awake`: `UIApplication.isIdleTimerDisabled` on the main thread + array entry.
- `android/jni/mob_nif.zig` — `nif_device_keep_awake` → `MobBridge.keepAwake(Int)` (field + export + `cacheOptional` + NIF table entry).
- `test/mob/device_test.exs` boolean-guard test, ADR, and the matrix `Idle timer` row flipped ❌ → ✅.

## Tests / checks

973 pass (1 new). `mix format`, `mix credo --strict`, `mix erlfmt --check src/`, `xcrun clang-format`, and `zig fmt --check` all clean.

## Merge coupling & verification

- Cross-repo: merge + release mob **before** an app regenerates its bridge (the generated Kotlin calls `MobBridge.keepAwake`). Same ordering as MOB-6 / MOB-15.
- Device verification pending but easy — "screen stays lit while enabled" is directly observable (and `adb shell dumpsys power` shows the wake state on Android).

### DIFF
--- a/android/jni/mob_nif.zig
+++ b/android/jni/mob_nif.zig
@@ -286,6 +286,9 @@ pub const BridgeMethods = extern struct {
     // MobBridge.orientationLock(Int) — calls activity.setRequestedOrientation.
     // Companion Kotlin method ships in the mob_new template (see PR notes).
     orientation_lock: jni.JMethodID = null,
+    // MobBridge.keepAwake(Int) — toggles the window's FLAG_KEEP_SCREEN_ON.
+    // Companion Kotlin method ships in the mob_new template.
+    keep_awake: jni.JMethodID = null,
     element_frames: jni.JMethodID = null,
     // ── Mob.Peripheral.VendorUsb ─────────────────────────────────────────
     // Each takes a pid as jlong (so Kotlin can echo it back when calling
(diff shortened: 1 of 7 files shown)


GenericJam:
Reviewed the code (read the full functions, not just grep) — **LGTM, endorse.**

- **iOS correct**: `nif_device_keep_awake` reads the bool atom then sets `[UIApplication sharedApplication].idleTimerDisabled = on` inside `dispatch_async(dispatch_get_main_queue())` — main-thread, as required for that UIApplication property.
- **FFI consistent**: zig caches `keepAwake` as `(I)V` and `CallStaticVoidMethod(..., on)`; matches the Kotlin `keepAwake(on: Int)` in the paired mob_new PR. `notLoaded` null-guard via `cacheOptional` gives the documented graceful drift (predates-bridge = no-op), same as `orientationLock`.
- **NIF wiring**: table entry + `src/mob_nif.erl` decl/stub present; the cross-layer `nif_declaration_test` (green in CI) covers the erl↔array consistency.
- ADR + `Idle timer` matrix row flip + a boolean-guard test included. Design (write-only, re-assert-on-resume, core `Mob.Device` not a plugin) is consistent with `lock_orientation/1`.

Only caveat: **not device-verified yet** (PR notes it as pending). Low-risk and trivially observable, but the other Device features (MOB-14 especially) merged device-verified. Happy to verify on the connected moto g (`dumpsys power` wake state) + iPhone SE before merge to hold the same bar — otherwise the code + tests are solid enough to merge.

PR #67, opened 2026-07-05
MOB-35: Mob.Audio input-level metering (agent "ears") — contract + host tests

GenericJam:
First stage of the agent-"ears" capability (MOB-35): a mic-level probe so an agent can detect whether the device is producing sound (e.g. loop *"keep going until you hear sound"*) — the audio analog of `screenshot`/`ui_tree`, and the iOS-capable answer that `mob_audio_capture` (Android output-mix only) can't provide.

**This PR — host-verifiable contract + foundation:**
- `Mob.Audio.start_input_metering/1`, `input_level/0`, `stop_input_metering/1`.
- `input_level/0` returns `{rms, peak} | :silent | {:error, reason}` — the **same shape** as `MobAudioCapture.output_level/0`, so `:mic` and `:output` read uniformly (the unified metering contract).
- NIF functions declared in `mob_nif.erl`; pure `decode_level/1` host-tested (25 pass). Format / erlfmt / credo --strict clean.
- Metering shares the mic session with recording (not split into a plugin).

**Not in this PR — next stage (needs device verification, per the on-device rule):**
- iOS native (`AVAudioRecorder` metering, reusing the existing recorder pattern) + Android native (`AudioRecord` bridge). The NIF funcs raise `not_loaded` until then.
- Device-verify the acoustic loop on the physical iPhone SE + Moto G (play audio → `input_level` rises → silent when it stops) — the mic-hears-speaker test.

### DIFF
--- a/android/jni/mob_nif.zig
+++ b/android/jni/mob_nif.zig
@@ -2691,6 +2691,39 @@ export fn nif_audio_stop_recording(
     return erts.ok(env);
 }
 
+// Audio input metering (mic level probe). Android impl (AudioRecord bridge) is
+// MOB-35 stage 3 — these stubs keep the NIF table consistent with iOS so the
+// module loads; input_level reports :not_implemented until the bridge lands.
+export fn nif_audio_start_input_metering(
+    env: ?*erts.ErlNifEnv,
+    argc: c_int,
+    argv: [*]const erts.ERL_NIF_TERM,
+) callconv(.c) erts.ERL_NIF_TERM {
+    _ = argc;
+    _ = argv;
+    return erts.ok(env);
+}
+
+export fn nif_audio_input_level(
+    env: ?*erts.ErlNifEnv,
+    argc: c_int,
+    argv: [*]const erts.ERL_NIF_TERM,
+) callconv(.c) erts.ERL_NIF_TERM {
+    _ = argc;
+    _ = argv;
+    return erts.atom(env, "not_implemented");
+}
+
+export fn nif_audio_stop_input_metering(
+    env: ?*erts.ErlNifEnv,
+    argc: c_int,
+    argv: [*]const erts.ERL_NIF_TERM,
+) callconv(.c) erts.ERL_NIF_TERM {
+    _ = argc;
+    _ = argv;
+    return erts.ok(env);
+}
+
 export fn nif_audio_play(
     env: ?*erts.ErlNifEnv,
     argc: c_int,
@@ -3732,6 +3765,9 @@ const nif_funcs = [_]erts.ErlNifFunc{
     .{ .name = "files_pick", .arity = 1, .fptr = nif_files_pick, .flags = 0 },
     .{ .name = "audio_start_recording", .arity = 1, .fptr = nif_audio_start_recording, .flags = 0 },
     .{ .name = "audio_stop_recording", .arity = 0, .fptr = nif_audio_stop_recording, .flags = 0 },
+    .{ .name = "audio_start_input_metering", .arity = 0, .fptr = nif_audio_start_input_metering, .flags = 0 },
+    .{ .name = "audio_input_level", .arity = 0, .fptr = nif_audio_input_level, .flags = 0 },
+    .{ .name = "audio_stop_input_metering", .arity = 0, .fptr = nif_audio_stop_input_metering, .flags = 0 },
     .{ .name = "audio_play", .arity = 2, .fptr = nif_audio_play, .flags = 0 },
     .{ .name = "audio_play_at", .arity = 3, .fptr = nif_audio_play_at, .flags = 0 },
     .{ .name = "audio_stop_playback", .arity = 0, .fptr = nif_audio_stop_playback, .flags = 0 },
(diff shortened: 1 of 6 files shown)


GenericJam:
Reviewed — **safe and a good foundation; one doc ask before merge, plus a merge-timing note.**

**The thing that looks like a blocker but isn't — I verified it.** Adding `audio_start_input_metering/0`, `audio_input_level/0`, `audio_stop_input_metering/0` to `-nifs([])` with no native symbol does **not** break `load_nif`. `-nifs([])` is an allowlist/upper-bound: `load_nif` requires every *native* symbol to be declared (the direction `nif_declaration_test` guards), but a declared entry with no native impl just keeps its Erlang stub. Confirmed by precedent — master already ships **98 `-nifs` declarations vs 81 android native symbols** (`battery_level`, all the `bt_*`, …) and runs fine. So this follows the existing pattern; `nif_declaration_test` correctly stays green.

**Good:**
- Unified contract — `{rms, peak} | :silent | {:error, reason}` matches `MobAudioCapture.output_level/0`, so `:mic` and `:output` read uniformly.
- `decode_level/1` extracted as a pure, host-tested kernel; the 25 tests cover passthrough, the `-120` dBFS `:silent` boundary, atom→error, and unknown→error.
- CI green, mergeable CLEAN, additive (no existing behavior touched).

**Ask (should-fix):** the docstrings for `start_input_metering/1` / `input_level/0` / `stop_input_metering/1` read as fully working, but on-device they **raise `not_loaded`** until the native follow-up — the NIF raises *before* `decode_level` can map it to `{:error, _}`, so a caller gets a raw crash, not a graceful error. Please add a "not yet implemented on-device — lands in the native follow-up; raises until then" note to those docs (and/or the moduledoc) so it's not a surprise. The `@spec … :: {float,float} | :silent | {:error,atom}` also doesn't match the interim raise.

**Merge-timing note:** this ships a public, documented API ahead of its implementation. Fine given it's additive, safe, and staged — but don't cut a mob release advertising audio "ears" until stage 2 (native) lands, or land them close together, so `input_level/0` isn't a dangling crash in a published version.

**Forward contract nit:** since `decode_level/1` derives `:silent` from a low-peak *tuple* and maps bare atoms to `{:error, _}`, the stage-2 native must return `{rms, peak}` for silence (not a bare `:silent` atom, which would become `{:error, :silent}`). Same shape as `output_level/0`, so likely already understood.

Net: mergeable; I'd add the not-yet-implemented docstring caveat first.

GenericJam:
## Review — approve with fixes (FFI seam verified clean end-to-end)

Traced the whole seam (`Mob.Audio` → `mob_nif.erl` arity → zig `nif_funcs`/JNI descriptor + `CallStatic{Void,Int}Method` → Kotlin `@JvmStatic` → iOS nif table). All arities/types/descriptors line up on both platforms — no runtime-crash risk on the call path, and the `mob_zig.zig` re-type of `CallStaticIntMethod` to a real fn-pointer is required + correct. Android is device-verified (moto g). Nice work.

Three things I'd fix before merge:

**1. The PR description is stale and will cause under-review.** The body says the iOS/Android native is *"Not in this PR — next stage … The NIF funcs raise `not_loaded` until then."* But the diff adds the full iOS impl (`ios/mob_nif.m`) and the Android zig NIFs (`android/jni/mob_nif.zig`, `mob_zig.zig`). ~120 lines of native code are actually here — please rewrite the body so reviewers don't skip them. (Also: the body says Android uses an `AudioRecord` bridge, but the mob_new #31 impl uses `MediaRecorder.getMaxAmplitude`.)

**2. `cacheRequired` makes this NOT drift-safe — boot-crash risk.** The three methods use `cacheRequired(jenv, "audio_input_level", "()I", …) → return -1`. If an app updates mob (this PR) but does **not** regenerate its `MobBridge.kt` from mob_new #31, `nifLoad` returns -1, the whole `mob_nif` fails to load, and the app **crashes at boot** (the classic mob_nif-declaration gotcha). Contrast the output-probes PR #54, which uses `cacheOptional` + null-guards and degrades gracefully. Please either (a) switch these to `cacheOptional` and no-op when the bridge lacks them, or (b) make the merge plan explicit: **mob_new #31 must land and existing apps must regenerate their bridge**, and never ship a mob with #67 against a bridge without #31.

**3. iOS metering is in the diff but neither compiled nor device-verified by CI.** CI here is host Elixir + `clang-format`/`swiftlint` (format-only) + mix_audit — there's no ObjC compile step and no device run. The code reuses the existing `AVAudioRecorder` pattern and looks right, but "looks right" is the whole gap. Please device-verify on the iPhone SE (the mic-hears-speaker loop) or gate iOS as experimental.

### Follow-ups (not blockers)
- **iOS no-permission blind spot:** without mic permission, iOS `record` returns silently and metering reads ≈ -160 dBFS → `:silent`, indistinguishable from genuine silence — an "ears" loop can't tell "denied" from "quiet" and may hang. Android surfaces a distinct error. Consider checking `AVAudioSession recordPermission` and returning a distinct atom.
- **`rms` semantics differ:** iOS returns `{averagePower, peak}` (true RMS-ish); Android returns `{db, db}` from `getMaxAmplitude` (a *peak*). Fine for boolean "is there sound?", wrong for numeric cross-platform comparison — worth a `@doc` note.
- Android's first `getMaxAmplitude()` after start returns 0 (a single-shot caller reads `:silent` once).

Merge the pair together; do **not** merge this without mob_new #31 given point 2.

_(Review verified against PR-head: FFI parity, the stale body, and `cacheRequired` are confirmed in the diff. The iOS-path items are flagged from reading, not device-tested.)_

PR #75, opened 2026-07-29
Fix vertical scroll width after iPad resize

dl-alexandre:
## Summary

Bind vertical scroll content to the live SwiftUI viewport width so rotating or resizing an iPad scene does not retain a portrait-sized content proposal. Horizontal scrolling remains unchanged.

## Validation

- Built and signed through a downstream Mob app on physical iPad12,1
- Verified portrait plus landscape-left and landscape-right
- Both 2160×1620 landscape orientations fill the complete canvas; the prior large black side margin is gone
- Cold terminate/relaunch still reconnects within two seconds with app data preserved
- Existing scroll keyboard-dismissal, accessibility identifier, background, padding, and observer behavior retained

No hard-coded device dimensions.


GenericJam:
Thanks for this — the mechanism is plausible and the merged tree compiles and passes the suite cleanly, but review turned up two blocking concerns before this can land:

1. **Intrinsic-sheet regression risk (new since your branch):** #94 (0.7.32) landed content-detent sheets that measure the sheet body under .fixedSize(horizontal: false, vertical: true). Your change wraps every vertical Scroll — on every device, not just iPad — in a GeometryReader, whose ideal height under an unspecified proposal is ~10pt (vs a ScrollView's content-sized ideal). A vertical Scroll inside a content-detent sheet would collapse to a hairline. This predates your branch point, so you couldn't have seen it — but it needs re-reasoning and a test against current master.

2. **Premise needs re-verification on master:** the stale-width symptom was validated against ~0.7.20; the scroll case itself is unchanged since then, but the surrounding layout pipeline moved 48+ commits (safe-area boot fix #74, fill_width row alignment #78, frame-registry #89/#90, intrinsic sheets #94). Could you confirm the black-margin repro still occurs on current master on your iPad12,1? If it does, a narrower fix — e.g. onGeometryChange-driven invalidation, or root-causing why the width proposal goes stale — would avoid putting a GeometryReader on the hot path of every vertical scroll.

Smaller notes for the next revision: the restructure drops two load-bearing comments (the Mob.Test scroll_info accessibility-id note and the iOS-17 scroll-observer caveat) and duplicates the modifier chain into both branches; if vertical scrolls retain stale width, horizontal scrolls presumably retain stale height (left untouched); and a CHANGELOG entry under [Unreleased] would be appreciated. We don't have iPad hardware in the automated pool, so your physical-device repro evidence genuinely matters here.

GenericJam:
Thanks for this, and sorry it sat so long.

First, the part that's on us: **this branch can no longer be rebased, and that's our fault.** We rewrote this repo's history after you forked (scrubbing a proprietary term), so your line carries ~293 commits that are pre-rewrite copies of our own commits. A trial merge produces add/add conflicts on files like `CHANGELOG.md`, `AGENTS.md` and `android/jni/*.zig` — files that exist on both sides but share no ancestry. Nothing you did caused that, and no amount of rebasing on your end fixes it.

Second, the good news: **three of the four changes in this branch are already on master.** Your comment text is there verbatim, so it looks like they landed via another route:

- button padding and background moved *inside* the `Button` label with `contentShape(Rectangle())` — `ios/MobRootView.swift:304-320`. This was a genuinely good catch; a padded area that looks tappable and isn't is a nasty bug.
- `case "baseline": return .lastTextBaseline` for row alignment — `:327`
- `maxHeight: node.fillHeight ? .infinity : nil` with `.topLeading` on column — `:314`

That leaves the scroll change, and here we got a result we didn't expect.

**We couldn't reproduce the iPad resize bug against a stock generated app, because a stock generated app never resizes.** `priv/templates/mob.new/ios/Info.plist.eex` declares no `UIDeviceFamily`, so a generated app installed on an iPad runs in iPhone compatibility mode. Measured on an iPad Pro 13-inch (M5) simulator with a native build:

```elixir
Mob.Test.screen_info(node)
#=> %{width: 320.0, height: 480.0, scale: 2.0, ...}
```

320x480 on a 13-inch iPad, letterboxed. `Mob.Device.lock_orientation(:landscape)` returns `:ok` and the viewport doesn't move.

So the symptom you're describing needs an app that actually resizes, which means your project differs from `mix mob.new` output somewhere — a modified `Info.plist`, a different build path, or something else. We'd genuinely like to know which, because it tells us what people are really doing with this.

**Tablets aren't currently in scope for mob.** That may change, and we've filed the gap so it isn't lost. If iPad support lands, this bug class becomes reachable and testable — `screen_info/1` and `element_frames/1` measure it precisely, which is how we checked the above.

I'm closing this rather than leaving it open indefinitely, but please do reopen or file fresh against current master if you can still reproduce it. The project has moved a long way since May (the renderer's child identity, lazy content and screen presentation have all been reworked), so a fresh reproduction would be worth more than a rebase of this one — and applied as-is today this diff would revert some of that, since it predates it.

Thanks again — your earlier docs PRs landed and this one found a real button bug on the way through.

GenericJam:
Closing as stale against current master — see above. Reopening or a fresh PR is very welcome if you can still reproduce on a recent version.

PR #76, opened 2026-08-08
docs: align screen lifecycle docs with single-GenServer navigation

minibikini:
The docs describe a per-screen-process model that the shipped implementation doesn't have, and `terminate/2` semantics that never fire the way they're described.

**What the code actually does** (verified against `lib/mob/screen.ex`):

- One `Mob.Screen` GenServer owns the whole navigation stack. The active screen is a `{module, socket}` pair in its state.
- `push` mounts the next module in that same process and stores `{module, socket}` in `nav_history`.
- `pop` / `pop_to` / `pop_to_root` / `reset` restore a snapshot from `nav_history` — they start no process and call no lifecycle callback.
- `terminate/2` runs only when the whole `Mob.Screen` GenServer stops (app exit, crash, shutdown), and delegates to the module active at that time only.

The native side confirms this is intentional: `mob_handle_back` (iOS `mob_nif.m`, Android `mob_nif.zig`) looks up a single registered `:mob_screen` process, and `renderer.ex` calls `set_root/1` once per render. There is no per-screen process anywhere.

**Docs fixed in this PR**

- `guides/screen_lifecycle.md` — "Each screen in the navigation stack is a separate, supervised process" and the "popped → terminate/2" diagram step.
- `guides/navigation.md` — "Passing data on pop" claimed popping restores a still-running per-screen process.
- `lib/mob/screen.ex` `@moduledoc` — claimed one supervised GenServer per screen with per-screen crash isolation.
- `PLAN.md` — "terminate/2 is called on every screen process exit" used for crash reporting.

No runtime code changes. Formatting + `mix credo --strict` are clean.

This is documentation-only. If a per-screen actor model is actually the intended direction, this PR highlights exactly where the shipped implementation and the docs diverge, and it would be worth deciding which way to take it before md docs keep promising isolate process-per-screen behaviour that doesn't exist.


GenericJam:
Triage note (draft, so no full review yet): the premise checks out — current Mob.Screen really is single-GenServer navigation ({module, socket, nav_history, render_mode}; push mounts into the same process, pop/reset restore snapshots without lifecycle callbacks), so this corrects real doc/code drift. However, a restructure toward per-screen processes is actively being explored, which would re-stale exactly the sections this PR fixes. Suggest holding further polish until that direction settles — we'll ping here when it does so you aren't diffing against a moving target. Thanks for the careful work either way.

GenericJam:
Thank you for this, and sorry it sat for so long.

You were right about all of it, and the audit did more than fix the docs — it forced the decision. You closed by saying that if a per-screen actor model was actually the intended direction, this PR showed exactly where the implementation and the docs diverged and it was worth deciding which way to go. That's what happened: rather than downgrade the docs to match a single-GenServer runtime, we built the runtime the docs described.

**MOB-112 (#99)** landed on 2026-08-29 and shipped in **0.7.33**. `Mob.Screen.Server` is now one process per live screen owning that screen's socket, started and monitored by an owner that holds the navigation state. **MOB-113 (#101)** then extracted that owner into `Mob.Router` — partly because, as its commit says, the moduledoc having to describe three jobs at once is a good part of why it had been wrong twice.

Every divergence you listed is now closed in the direction you asked about:

- **Per-screen processes** — `Mob.Router.start_screen/5` calls `Mob.Screen.Server.start_link` and keys live screens by pid.
- **Crash isolation** — a crash in one screen's `handle_event` no longer takes navigation and its siblings with it. The owner observes the exit, restarts that screen and repaints, capped at 5 restarts in 10s so a screen that crashes on every render can't spin.
- **`terminate/2`** — fires per screen now. `Mob.Screen.Server.terminate/2` delegates to your module, and the owner's `stop_screen/2` runs it on pop, not only on app exit.
- **`guides/screen_lifecycle.md`, `guides/navigation.md` and the `Mob.Screen` moduledoc** all now describe what the code does.

The commit message for #99 credits this PR directly — "mob#76 corrected the docs, which documented the gap rather than closing it" — and so does the 0.7.33 changelog entry: *"`Mob.Screen`'s moduledoc had claimed this for a long time and mob#76 had to correct it; the claim is now true."*

Closing this one only because it can't merge as it stands: it was branched from a point whose history has since diverged, so the diff carries the whole tree rather than the four files you meant to touch. That's a trap on this repo, not anything you did — it caught me on my own changelog PR this week too.

The underlying issue is resolved and you're the reason. Please do send more like this.

GenericJam:
Resolved by MOB-112 (#99) in 0.7.33 — the per-screen process model this PR identified as missing. See the comment above.

minibikini:
thanks!

PR #77, opened 2026-08-10
fix: cold-start launch notifications dropped before nif_load (Android + iOS)

asheehan:
Both platform stores for the launch notification bail out when called before `nif_load` has created their mutex:

- Android: `mob_set_launch_notification` — `const mutex = g_launch_notif_mutex orelse return;`
- iOS: `mob_set_launch_notification_json` — `if (!g_launch_notif_mutex) return;`

But the mutex is created in `nif_load`, and the cold-start path runs strictly **before** that: on a notification tap that launches a killed app, `MainActivity.onCreate` (Android) / the app delegate (iOS) store the payload before the BEAM thread boots. The payload is silently discarded every time — **tap-to-open from a killed app has never worked on either platform**. Warm/backgrounded taps (`onNewIntent` / running delegate) are unaffected, which makes the bug easy to miss in testing.

The fix mirrors `mob_set_opened_document`, which has always stored pre-mutex and documents why it's safe: nothing reads the global until `take_launch_notification`, which can only run after `nif_load`, so the unguarded window has no concurrent reader. Post-load stores still lock. (`zig ast-check` passes; the objc change is the same two-line guard restructure.)

Found while wiring FCM push in a mob app (BuyFrost): an adversarial review noticed the opened-document sibling's comment describes exactly this cold-launch window, exposing the launch-notification variant as a latent bug rather than a design choice.

CHANGELOG entry added under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,21 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- **Cold-start launch notifications are no longer dropped (Android + iOS).**
+  `mob_set_launch_notification` (Android) and `mob_set_launch_notification_json`
+  (iOS) bailed out when called before `nif_load` had created their mutex — which
+  is exactly when the cold-start path runs (MainActivity.onCreate / the app
+  delegate store the tapped notification before the BEAM boots). The payload
+  was silently discarded, so tap-to-open from a killed app never worked; a
+  warm/backgrounded tap (delivered via `onNewIntent` / the running delegate)
+  was unaffected. Both setters now store before the mutex exists, the same
+  pre-mutex pattern `mob_set_opened_document` has always used — safe because
+  nothing reads the global until `take_launch_notification`, which can only
+  run post-`nif_load`.
+
 ## [0.7.20] - 2026-07-11
 
 ### Changed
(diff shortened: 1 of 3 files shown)


asheehan:
Filed #81 to track the underlying bug.

GenericJam:
Reviewed, verified, and landed — thank you for an excellent fix and the precise diagnosis in #81.

The review confirmed everything: both setters bailed pre-nif_load on master, the cold-start ordering means the unguarded pre-mutex write has no concurrent reader (onCreate stores before the BEAM thread spawns; the only reader can't run before nif_load), and it's byte-for-byte the accepted mob_set_opened_document pattern. We A/B-verified on an Android emulator: baseline NIF drops the payload on a true cold start (take_launch_notification → :none, confirmed down to the early-return in the disassembly); with your fix the same cold start delivers the payload end-to-end into a screen's handle_info. 1280 tests pass on the rebased tree.

Your branch had drifted 48 commits (only CHANGELOG conflicted), and pushing a rebase back to your fork wasn't possible from this side, so your commit was rebased and pushed to master directly with your authorship preserved: ed1e275. Closing this PR as landed — the fix ships in the next release. One tiny follow-up we're tracking separately (pre-existing pattern, also in mob_set_opened_document): the setters read the mutex global twice, so capturing it once into a local would close a vanishingly narrow TOCTOU window across all four functions.

PR #80, opened 2026-08-11
iOS harness: stop tap_xy reporting phantom successes; add pixel sampling for colour

GenericJam:
The on-device test harness reported success for things that did not happen, and could not see colour at all. Both made agent-driven verification on iOS unreliable — an iOS renderer bug sat unverified in a downstream repo for weeks partly because of this.

## 1. `tap_xy` reported `:ok` when nothing happened

On a **physical iPhone** it returned `:ok` for every coordinate — including empty space — and nothing ever happened; `mob_send_touch_phase` returns `YES` for API *availability*, never delivery. On the **simulator**, tapping a Box with `on_tap:` returned `:ok` while the handler never ran, because SwiftUI gives a plain `.onTapGesture` no accessibility action.

Now the return value reflects whether a handler actually ran: a process-wide atomic counter is bumped by every send helper that routes a user event into the BEAM (`mob_send_tap`, `mob_send_event`, `mob_send_change`), sampled before injection and polled for 300 ms after.

| Value | When |
|---|---|
| `:ok` | an event reached the BEAM within 300 ms |
| `{:error, :no_view_at_point}` | hit-test found nothing (now on both platforms) |
| `{:error, :no_effect}` | input accepted, no handler ran |
| `{:error, :no_element_at_point}` | simulator: view present, no AX element |

`@doc` now states the real per-platform capability, including that **coordinate tapping does not work on a physical device** and that the counter only sees Mob's own handlers (sidecar caveat). `tap_xy` also moved to `ERL_NIF_DIRTY_JOB_IO_BOUND` — it was already sleeping on a normal scheduler.

Device-verified: a theme chip (Box + `on_tap`) now returns `{:error, :no_effect}` where it previously returned a false `:ok`.

## 2. `view_tree` colour: measured, then abandoned as designed

Reading `UIView.backgroundColor`/`UILabel.textColor` gave colour for **2 of 443** nodes. Also walking `CALayer`/`CAShapeLayer`/`CATextLayer` reached **4 of 443**. The new `ui_paint_debug/0` census says why:

```
204x SwiftUI._UIInheritedView / CALayer            all paint props 0
 64x SwiftUI._UIInheritedView / SwiftUI.SDFLayer   all paint props 0
 14x _UIInheritedView / CGDrawingLayer             all 0, contents=14
```

SwiftUI on iOS 26 paints via `SDFLayer` or rasterises into `contents`, exposing no readable paint property. Layer introspection is a dead end for SwiftUI content — recorded in an ADR rather than iterated on further. The colour fields stay (correct for UIKit chrome, root background, future Android) alongside a new `:class` field so a caller can tell *which renderer* drew a node and therefore why a colour is `nil`.

## 3. `sample_color/2` — colour verification that works

`sample_region/4` returns raw RGBA for a **natively cropped region** (small payload over distribution), reduced by a pure, unit-tested Elixir function to `average`, `dominant`, `dominant_share` and `distinct` — because a card with text on it is not one flat colour, and the share/distinct fields tell you whether to trust `dominant`.

Gated **strictly** `#if !MOB_RELEASE`, deliberately *not* the `|| defined(MOB_ENABLE_SCREENSHOT)` its neighbour uses: arbitrary-rect pixel reads let a caller reconstruct the screen region by region, which would hand release builds the capability that opt-in exists to gate. Pinned by a test on the guard string.

**Acceptance test, on device:** two chips, `background: :primary` vs `:surface_raised`, under a glass theme.

| Build | active | inactive | |
|---|---|---|---|
| without the glass fix | `0xFF20202E` | `0xFF20202E` | bug reproduced |
| with #78 | `0xFF4000AF` | `0xFF101023` | fix confirmed |

The regression that previously required hand pixel-diffing is now `assert a.dominant != b.dominant`.

## Also reported, not fixed

`scroll_info`, `scroll_to` and `element_frames` are behind `#if !MOB_RELEASE` on iOS while Android registers them unconditionally, so an iOS release build raises `nif_error(not_loaded)` and kills the calling screen process. None of the three touches private API — the `screenshot` opt-in pattern would fit.

## Verification

1011 tests passing, `format`/`credo --strict`/`erlfmt`/`clang-format` clean; ObjC compiled and run on an iPhone 17 simulator; non-vacuity of new tests proven by sabotage-and-restore.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,19 @@ Mob.Test.inspect(node)           # full snapshot: screen, assigns, nav stack, wi
 This is faster, exact (not pixel-inferred), and works without taking a
 screenshot. Use it as the default.
 
+**Colour is the exception.** `view_tree/1`'s `bg_color`/`text_color` are `nil`
+for virtually all SwiftUI content (iOS 26 paints via `SDFLayer` or rasterises
+into `contents` — colour for 4 of 443 nodes when measured). To verify what was
+actually drawn, sample pixels:
+
+```elixir
+Mob.Test.sample_color(node, "my-card")   # {:ok, %{average: 0xFF2196F3, dominant: ..., ...}}
+```
+
+It crops in the NIF, so only that element's pixels cross dist. See
+`decisions/2026-08-09-view-tree-colour-needs-screenshot-sampling.md` and
+`decisions/2026-08-10-sample-region-crops-natively-and-stays-debug-only.md`.
+
 ### Drive
 
 ```elixir
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -212,6 +212,17 @@ These are the things we've burned ourselves on. Following them isn't optional.
     rather than duplicating them. See
     `mob_dev/decisions/2026-06-19-mob-adopt-lives-in-mob_dev.md`.
 
+14. **Don't drive iOS by coordinate — use `Mob.Test.tap/2`.** `tap_xy/3` works
+    on the simulator only for elements SwiftUI gives an accessibility action
+    (`Button`, text fields); a `Box` with `on_tap:` has none. On a physical
+    device the injected IOHID touch is accepted and never delivered, so every
+    coordinate fails. Both now return `{:error, :no_effect}` instead of the
+    `:ok` they used to — a harness that reported success for taps that did
+    nothing let a downstream iOS renderer bug sit unverified for weeks. When
+    you add a harness NIF, make its success value mean *observed effect*, never
+    *the platform API didn't complain*. See
+    `decisions/2026-08-09-tap-xy-reports-observed-effect.md`.
+
 ## Where to look
 
 | Question | File |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -223,13 +223,23 @@ over Erlang distribution to a remote test runner or agent.
   `[{type, label, value, {x,y,w,h}}, ...]` tuples. Works on any app with zero modification.
 - `ui_debug/0` — raw accessibility dump for debugging
 
-**Phase 2 — Synthetic interaction (complete)**
+**Phase 2 — Synthetic interaction (partial — see the honesty note below)**
 
 - `tap/1` — tap by accessibility label
 - `tap_xy/2` — tap at screen coordinates (with responder-chain walk to focus text fields)
 - `type_text/1` — type into the focused text field
 - `delete_backward/0`, `key_press/1`, `clear_text/0` — keyboard control
 - `long_press_xy/3`, `swipe_xy/4` — gesture synthesis
+
+Coordinate-driven input is *not* finished, whatever the list above implies.
+`tap_xy/2` now returns `:ok` only when the app demonstrably reacted; on the
+simulator that limits it to `Button`s and text fields, and on a physical device
+the injected IOHID touch is accepted but never delivered, so every coordinate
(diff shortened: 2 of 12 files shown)


GenericJam:
Adversarial review + on-device verification complete. **Verdict: approve the substance — rebase required before merge.**

The branch forked at 0.7.20; master has 15 commits touching `ios/mob_nif.m` since. Merge-tree shows exactly one conflict hunk, in `nif_tap_xy`: master's side is MOB-99's retry-hardened atomic tap (`mob_retry_main_thread_bool` + `find_a11y_at_point_in_current_windows`); this branch still has the pre-MOB-99 two-`dispatch_sync` shape. Taking this side verbatim would regress MOB-99 — the counter/hit-test-first/await logic needs folding into MOB-99's atomic block. Everything else auto-merges.

The substance is NOT superseded: verified on current master that the device path still returns `:ok` off API availability and the sim path still returns `:ok` on `accessibilityActivate` with no effect check — the phantom-success bug this fixes is still live. Rebasing also fixes the id-addressed sampler path for free (reproduced the MOB-102 `element_frames` staleness on this branch on-device; rect-addressed sampling unaffected).

On-device (sim, 1011/1011 tests + all format/lint gates green on the branch): miss → `{:error, :no_effect}`; outside windows → `{:error, :no_view_at_point}`; real button → `:ok` with assigns actually changing; Box with `on_tap` → honest `{:error, :no_effect}` (previously phantom `:ok`). Pixel sampler round-trips sRGB token hexes bit-exact through P3/16-bit capture (Onyx `0xFF282424`, Ivory `0xFFFFFAF4`, share 1.0) — this can replace the downstream token-audit's 16-bit-unsupported bailout outright.

Two non-blocking notes for the rebase: (1) `g_ui_event_seq` is process-wide and scroll notifications bump it, so an unrelated event in the 300ms settle window can false-positive — fine for a serial harness, worth a doc sentence; (2) post-#94 the hard 'Box has no AX element' wording should be retested/softened (#94 added `.isButton` but no `accessibilityAction`, so the claim likely still holds). Nit: unchecked `enif_alloc_binary` returns in `nif_sample_region`/`nif_ui_paint_debug`.

GenericJam:
Rebased onto master (848eabd) with the MOB-99 fold done as planned: MOB-99's atomic retry skeleton kept verbatim (mob_retry_main_thread_bool, find+act in the same window, text-field responder walk), with the honesty logic folded in — seq snapshot before injection, hit-test-first {:error, :no_view_at_point}, post-activation settle wait → {:error, :no_effect}. Retry only re-runs find/activate, so no double-fire; settle runs once after a successful activation.

Re-verified on the rebased HEAD (4a88f6f): 1276 tests / 0 failures, format, credo --strict, warnings-as-errors, erlfmt, clang-format all clean. Device matrix re-run on a pool sim: honest miss/out-of-window errors, real button :ok with assigns advancing, 5 rapid taps all :ok (MOB-99 intact), tap-during-re-render :ok, sampler bit-exact (0xFF282424 / 0xFFFFAF4 wait — 0xFFFFFAF4, share 1.0), id-addressed sample_color stable across 3 forced re-renders (MOB-102 fix confirmed working with this branch).

Also included a review-fixes commit: enif_alloc_binary returns checked in nif_sample_region/nif_ui_paint_debug; process-wide-counter caveat documented beside the sidecar caveat (scroll events bump the counter — serial-harness assumption); and the post-#94 Box prose corrected to the device-verified truth: a role-button Box IS a findable AX element with .isButton but exposes no accessibilityAction, so activation is a verified no-op → honest {:error, :no_effect}.

Merging.

PR #85, opened 2026-08-26
Fix component handle pool exhaustion crashing the screen process (MOB-100)

GenericJam:
## Summary
- Fixes a downstream-reported bug: a screen registering ~60 native components (icon catalog) crashed its `:mob_screen` process with `badarg` from `register_component`, going tap-dead until force-kill.
- Three compounding defects, all fixed:
  1. **Graceful exhaustion** — full pool returned the same `badarg` as a malformed pid, crashing `Mob.ComponentServer.init` and the whole screen. Both native bridges now return `{:error, :component_slots_exhausted}`; the component fails individually, the screen survives.
  2. **Slot-0 leak** — slot 0 (a valid pool index) was conflated with the `:no_render` sentinel (also 0), so `terminate/2`'s `if handle != 0` guard never deregistered it. Unified sentinel is now `-1`.
  3. **Root cause found while writing the regression test**: `Mob.ComponentRegistry.reconcile/2` stops components via `Process.exit(pid, :shutdown)`, but `Mob.ComponentServer` never trapped exits — `terminate/2` (and its `deregister_component` call) never ran for *any* component leaving a screen's tree, not just slot 0. This was the dominant leak. See `decisions/2026-08-26-component-pool-trap-exit.md`.
  4. Bumped `MAX_COMPONENT_HANDLES` 64 → 256 on both platforms as headroom (still fixed-size; growable pool tracked as a follow-up).

## Test plan
- [x] `mix test` — 1034 passed, including new regression coverage in `test/mob/component_server_test.exs` (graceful exhaustion, slot-0 reuse, and a test exercising the *real* `Mob.ComponentRegistry.reconcile/2` stop path to prove no leak)
- [x] `mix format` / `mix credo --strict` / `mix erlfmt --check src/` / `clang-format` / `swiftlint` — all clean
- [x] Device-verified on a physical Android phone (moto g power) and the iOS simulator: mounted 300 native components (44 over the old 64 cap, 44 over even the new headroom), screen stayed fully responsive (`taps` counter kept incrementing), allocated exactly 255 unique handles bounded within the pool, unmount/remount cycling reused the identical slot range (max 255) with zero monotonic growth.

Linear: MOB-100

### DIFF
--- a/android/jni/mob_nif.zig
+++ b/android/jni/mob_nif.zig
@@ -944,7 +944,12 @@ export fn nif_swipe_xy(
 // mob_nif_init_state (called from mob_nif.c's nif_load BEAM callback).
 
 const MAX_TAP_HANDLES: usize = 256;
-const MAX_COMPONENT_HANDLES: usize = 64;
+// MOB-100: bumped from 64 — a single screen legitimately rendering ~60
+// components (e.g. an icon catalog) plus a few leftover slots from prior
+// navigation could tip over the old cap. Keep in sync with the identical
+// constant in ios/mob_nif.m. Still fixed-size: a growable pool or component
+// recycling is a longer-term follow-up, not this fix.
+const MAX_COMPONENT_HANDLES: usize = 256;
 
 /// Per-tap slot: the registered pid, an optional caller-supplied tag, and
 /// the throttle state for high-frequency events. tag_env is non-null while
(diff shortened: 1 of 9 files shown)


GenericJam:
Both findings verified and fixed in cd4b59f:

1. **Version-skew catch-all**: `register_native_handle/4` now handles a stale native binary's bare-int success return and raising `badarg`/`ArgumentError` on exhaustion (both the pre-fix contract), degrading to the `-1` sentinel + a logged warning instead of crashing. Covered by two new tests simulating each legacy shape.

2. **Sibling tap-pool exhaustion**: `nif_register_tap` on both platforms now returns `-1` instead of `badarg` when the 256-slot tap pool is full. Turned out to need zero `Mob.Renderer` changes — every `mob_send_*` sender already no-ops on an out-of-range handle, so the existing bare-int contract already accommodates a sentinel. Device-verified on physical Android + iOS simulator: a screen with 301 tappable buttons (45 over cap) boots and stays fully responsive; the native "pool exhausted" log fires repeatedly with zero crashes.

Full details in the commit message.

PR #90, opened 2026-08-27
Bump to 0.7.31 — frame-tracker ownership fix + honest platform docs

GenericJam:
First run of the release gate that landed in 0.7.30. Reviewing `0.7.29..master` caught three things in the MOB-103 work **that 0.7.30 already shipped**.

## 1. iOS frame tracker lost its ownership token on a refused write (real bug)

`record` assigned `box.seq` unconditionally, so a generation-refused write (returns 0) overwrote a valid seq with 0 — and `mob_unregister_frame` ignores seq 0, permanently disabling that tracker's own cleanup.

Failure: push from a screen tagging `:id` `"save"` to one whose `"save"` is a lazy row below the fold. The outgoing screen's `.move` writes are refused by the generation gate, `box.seq` becomes 0, its `.onDisappear` is a no-op, the id stays in the tree so the purge keeps it — and `Mob.Test.tap_id` taps the **old screen's coordinates**.

Keeping the last *successful* seq is strictly safe: the compare-and-delete still refuses to delete whenever an incoming tracker has claimed the id since. The `.opacity` (reset) transition doesn't hit this, which is the tell — it doesn't move the outgoing frame, so nothing gets refused.

## 2. `element_frames/1` docs implied cross-platform parity that doesn't exist

The drop-when-not-laid-out behaviour is **iOS only**. Android clears frames wholesale on a navigation transition and never per element (`MobBridge.setRootJson`), so a scrolled-away `lazy_list` row keeps its frame and `tap_id` happily taps it. Verified in the template source. Docs now say so plainly instead of implying parity; Android parity wants its own issue.

## 3. 0.7.30 published with empty release notes

CHANGELOG's top section was still `## [Unreleased]` when 0.7.30 was tagged, so `release.yml`'s `## [<version>]` extraction found nothing. Retitled to `[0.7.30]` so the notes exist for anyone reading back. Exactly the version/release-seam class the gate exists to catch.

## Also

Wires `make -C test/native run` into the macOS CI job. That harness is the only automated guard on the registry's ownership/generation/purge rules, and it *reproduces* the algorithm rather than linking it — running it is what keeps the copy honest. It was committed in 0.7.30 but nothing ran it.

## Preflight

`mix format --check-formatted`, `mix credo --strict`, `mix compile --warnings-as-errors`, 1111 tests, `clang-format`, and the native harness (20/20) all clean.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,6 +76,14 @@ jobs:
             ios/mob_nif.m \
             android/jni/mob_beam.h
 
+      - name: Native harness (frame registry — MOB-102/MOB-103)
+        # The only automated guard on the element-frame registry's ownership,
+        # generation-gating and purge rules. It reproduces the algorithm rather
+        # than linking ios/mob_nif.m (which needs UIKit + the swiftc-generated
+        # header), so it can drift — running it in CI is what keeps the copy
+        # honest. Same toolchain the clang-format step above already uses.
+        run: make -C test/native run
+
       - name: Install swiftlint
         run: brew install swiftlint
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,26 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
 
 ---
 
+## [Unreleased]
+
 ## [0.7.31] - 2026-08-27
+
+### Added
+- **`mob_send_dismiss` (Android)** — sends `{:dismiss, tag}` to the process
+  registered for a handle, the shape `Mob.UI.sheet/2` documents for
+  `:on_dismiss` and the one iOS has always delivered. Android had no dismiss
+  sender at all, so generated sheet renderers fell back to the tap sender and
+  delivered `{:tap, tag}`; a screen written to the documented contract never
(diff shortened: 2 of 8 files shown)


GenericJam:
## `mob_send_dismiss` device-verified (physical Moto G power, 2021)

Verified through the full JNI chain rather than just statically:

`onDismissRequest` → `MobBridge.nativeSendDismiss` → `Java_com_example_sheetprobe_MobBridge_nativeSendDismiss` → **`mob_send_dismiss`** → `sendEvent(handle, "dismiss")` → `{:dismiss, tag}` delivered to the screen's `handle_info/2`.

The probe screen matched both `{:dismiss, tag}` and `{:tap, tag}` so a wrong shape would be logged rather than kill the screen. Result across two open/dismiss cycles: `log: [:dismiss, :dismiss]`, the `:tap` clause never fired, screen alive throughout, sheet re-presented correctly after each dismissal.

Paired with GenericJam/mob_new#38, which supplies the `nativeSendDismiss` extern and JNI thunk.

**Release ordering:** mob_new's CI clones `GenericJam/mob` at master with no ref pinned, so merging this PR unblocks mob_new's CI immediately — it currently fails with `call to undeclared function 'mob_send_dismiss'`, which is the lockstep working as intended.

Issue #39, opened 2026-05-29
x86_64 Android - Build Script


andyl:
In #20, I reported that on my host (Xubuntu 24.04), I am not able to run a Mob app in the Android emulator, because Mob does not distribute an x86_64 Android Beam Runtime (the Erlang VM and standard libraries).   

Here I publish a bash script `build-otp-android-x86_64.sh` (see attached) that builds the missing runtime.

[build-otp-android-x86_64.sh](https://github.com/user-attachments/files/28371854/build-otp-android-x86_64.sh)

With the generated runtime, my Mob application runs in the emulator, with certain features broken, and periodic device freezes. 

Still, this seems to move the ball in the right direction, so I am posting my current WIP.  

To use the x86 runtime in the emulator:

```
> export MOB_ANDROID_OTP_x86_64="$HOME/.mob/x86_64-build/out/otp-android-x86_64" 
> mix mob.deploy --native 
```

GenericJam:
Resolved by the official x86_64 Android support that landed for #20.

You no longer need a side-loaded build script or the `MOB_ANDROID_OTP_x86_64` env override: `otp-android-x86_64-7d46fdd4.tar.gz` is published on the `otp-7d46fdd4` release, and **mob_dev 0.6.4** downloads it automatically (`OtpDownloader.ensure_android("x86_64")`) and builds the x86_64 JNI libs. `mix mob.new` apps (**mob_new 0.4.5**) ship the `x86_64` abiFilter, so `mix mob.deploy --native` just works on an x86_64 emulator.

The official build scripts (`scripts/release/*x86_64*` in mob_dev) also fix the freezes/broken features you saw with the WIP runtime: the root cause was an incomplete xcomp conf (missing AR/RANLIB, gc-sections, the removed ld.gold). Thanks for the WIP script that moved this forward. Closing as superseded by #20.

PR #94, opened 2026-08-27
feat(ui): add intrinsic sheets and box semantics

GenericJam:
## Summary

- add intrinsic-content Sheet detents, including an optional maximum height
- validate and encode detents at the renderer boundary so raw/sigil nodes cannot bypass the contract
- measure, cap, re-clamp, and internally scroll intrinsic Sheet content on iOS
- expose generic Box accessibility labels, button roles, and disabled semantics
- include Sheet in both packaged platform tag manifests

## Why this belongs in Mob

These are generic renderer capabilities used by composed UI libraries. A plugin can style and compose `Mob.UI.sheet/2`, but it cannot change how the core Sheet measures arbitrary Mob children or make the core Box renderer honor accessibility props without replacing those primitives with app-specific native leaves.

The coordinated generated-Android implementation is [GenericJam/mob_new#39](https://github.com/GenericJam/mob_new/pull/39).

## Verification

- `mix credo --strict`
- `mix test --exclude macos_only --exclude requires_zig` — 1,119 passed, 38 excluded
- `git diff --check origin/master..HEAD`

This PR intentionally does not bump or publish a package version. Release bookkeeping follows after upstream review.

### DIFF
--- a/ios/MobNode.h
+++ b/ios/MobNode.h
@@ -235,8 +235,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, strong, nullable)
     NSDictionary *nativeViewProps; // full props dict forwarded to the factory
 
-// Accessibility — set from the tap tag atom name; read by XCTest / ui_describe_all
+// Accessibility — identifiers support test addressing; labels and disabled
+// state describe composite controls such as tappable boxes.
 @property(nonatomic, copy, nullable) NSString *accessibilityId;
+@property(nonatomic, copy, nullable) NSString *accessibilityLabel;
+@property(nonatomic, copy, nullable) NSString *accessibilityRole;
+@property(nonatomic) BOOL disabled;
 
 // Icon — logical name resolved to an SF Symbol on iOS / Material Symbol
 // on Android. textSize and textColor control glyph sizing + tint.
@@ -271,7 +275,7 @@ NS_ASSUME_NONNULL_BEGIN
 // system-default box in practice), but a sheet's corners are visibly
 // square-vs-rounded, so `sheetCornerRadius` gets its own -1 sentinel
 // (unset — use the system default) instead. `sheetDetents` is the raw
-// "medium"/"large" string list from Mob.Renderer — mapped to
+// built-in strings or content-detent dictionaries from Mob.Renderer — mapped to
 // PresentationDetent by MobSheetView (see MobRootView.swift), not here, so
 // this header stays framework-agnostic. Indicator geometry defaults to -1
 // (unset — use the system default indicator); Mob.UI.sheet's validation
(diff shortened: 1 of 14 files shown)


GenericJam:
## Adversarial review addressed — `34866b5`

Ran a high-effort adversarial pass over `894bf5d`. Eight findings, plus two of my own. All fixed on this branch.

### iOS Box — three interacting fixes

**`.allowsHitTesting(false)` let taps fall through a disabled box.** It makes a view *transparent* to touches rather than blocking them, so a `Box` with `disabled: true` used as a blocking overlay or dimmed backdrop passed taps straight through to the content it was meant to shield. Now `.disabled(node.disabled)`, which blocks interaction in the subtree while still consuming the touch.

**Toggling `disabled` destroyed the subtree's SwiftUI identity.** `ifLet` is `@ViewBuilder` if/else — `_ConditionalContent` — so flipping the branch hands SwiftUI a structurally different view and tears the subtree down. A `Box` wrapping a `TextField` lost in-flight text and focus; one wrapping a `Sheet` re-presented it. Tap wiring now branches on handler presence only, with the disabled check inside the closure, and both traits are unconditional OptionSet modifiers (`? .isButton : []`).

**Traits leaked to descendants without a label.** `.accessibilityAddTraits` is only scoped when `.accessibilityElement(children: .ignore)` ran, which only happened inside the label branch — so a role-only box announced every nested `Text` as its own button. Collapsing now keys on label **or** explicit button role.

### iOS Sheet

**Content detents flashed a 1pt hairline on every presentation.** Content can only be measured after the sheet is up, so the first detent came from sentinels (`intrinsicContentHeight = 1`, environment default `1`) and resolved to `.height(1)` — on every `MobSheetView`, not just the first. Measurement is now `Optional`; unmeasured presents at `.medium` and switches to the exact height when the measurement lands. The environment default is `0` meaning "root geometry unknown", so it no longer doubles as a clamp.

**The detent ignored the bottom safe-area inset.** `.height()` sets *total* sheet height while the content region is inset by the home indicator, so an intrinsic sheet was ~34pt short on notched devices — last rows under the indicator, and scrollable when the content was meant to fit exactly. The preference now carries height and inset together.

**The `0.9` ceiling is a named constant** with its rationale.

### Elixir

**Validation messages regressed** to one generic string with no `got:`. Restored per-rule messages — `[:medium, :medium]` says "must not contain duplicates, got: …" — plus a dedicated message for malformed `%{type: :content}` maps, which previously fell through to the built-in-subset branch and misdescribed the problem.

**`@external_resource` added** to both tag manifests in `Mob.Sigil` and `Mob.ScreenCase`. They're read into module attributes at compile time, so adding `Sheet` recompiled nothing for anyone with a warm `_build` — every path-dep consumer, i.e. the standard mob dev setup — leaving the sigil rejecting `<Sheet>` until a manual `mix clean`. This change is the first to actually trip it.

**Documented two limitations** the implementation implies: a scrollable child (`scroll`, `lazy_list`) reports full content height under `.fixedSize`, so it expands rather than scrolling independently; and invalid detents now *raise* at the renderer boundary rather than degrading to `[:medium, :large]`.

### On the native tests

They broke on **four separate behaviour-preserving refactors** here — naming a constant, unwrapping an Optional, adding the inset, making traits unconditional. They pinned source text, not behaviour. Retargeted at the mechanisms (the unmeasured fallback, the inset arithmetic, the non-branching traits, `.disabled` over `.allowsHitTesting`) so they break when behaviour regresses rather than when the code is tidied.

### Verification

1119 tests, `mix format`, `credo --strict`, `clang-format`, `swiftlint` (only the pre-existing `force_cast`), and `clang -fsyntax-only` on `mob_nif.m` — all clean. Device verification of the sheet still to come.

PR #104, opened 2026-08-30
fix(ios): honor fixed-height boxes

GenericJam:
## Summary

- apply `fixedHeight` when a MobBox has no fixed width
- preserve the existing full-width behavior for width-less boxes
- add a focused native source-contract regression

## Verification

- `mix format`
- `mix credo --strict`
- `mix test` (1,259 passed, 38 excluded)
- native iOS simulator build/install on pool-ios-3
- `Mob.Test.frame/2`: height-120 box `{0.0, 62.0, 402.0, 120.0}`; following height-40 reference `{0.0, 182.0, 402.0, 40.0}`

### DIFF
--- a/ios/MobRootView.swift
+++ b/ios/MobRootView.swift
@@ -767,6 +767,10 @@ private struct MobBox: View {
                     height: node.fixedHeight > 0 ? CGFloat(node.fixedHeight) : nil,
                     alignment: alignment
                 )
+            } else if node.fixedHeight > 0 {
+                stack
+                    .frame(height: CGFloat(node.fixedHeight), alignment: alignment)
+                    .frame(maxWidth: .infinity, alignment: alignment)
             } else if node.fillHeight {
                 // fill_height: true is what lets a wrapping box stretch to the
                 // viewport so center alignment lands on the visible midpoint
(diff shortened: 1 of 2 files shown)


GenericJam:
Adversarially reviewed and device-verified. **Approve.**

Device matrix on a pool sim (element_frames, points): height-only 120 → {402.0 × 120.0} exact with full parent width per contract — bug fixed; width-only, both-dims, and neither all unchanged; weighted box with fixed height → weight wins (269 = 538/2), matching Android's Compose semantics where weight's exact constraints coerce height(). Paint verified by pixel sampling: the height-only box's full 402×120 frame samples the exact requested color at dominant_share 1.0 (434,160 px @3x) — background/border/tap cover the expanded frame because the new .frame sits before padding/background/gestures in the chain, same ordering as Android. Gates on the PR+master merge (clean merge-tree): 1281 tests / 0 failures, format, credo --strict, warnings-as-errors.

Non-blocking: (1) no CHANGELOG entry — I'll include one in the release cut; (2) the new source-contract test asserts an exact 5-line whitespace-sensitive block — the #98 precedent is token-level =~ assertions plus an ordering check, worth loosening next touch; (3) precedence corner: height + fill_height together now lets height win on iOS while Android's fillMaxHeight().height(h) lets fill win — iOS is now internally consistent, but the contradictory-props corner diverges cross-platform; deserves a comment or docs note. Also discovered during review (filing separately): iOS Column/Row still ignore fixed width/height entirely — the same bug class this fixes for Box.

Merging.

PR #108, opened 2026-08-30
fix(theme): avoid repeated host NIF probes

GenericJam:
## Summary
- serialize and memoize the first Mob.Theme.color_scheme/0 NIF availability probe
- return the documented light fallback without retrying a failed host on_load on every call
- detect a later successful native code load so hot deploys recover the native color-scheme path

## Validation
- strict regression: repeated sequential and concurrent host reads
- full suite: 1281 passed, 38 excluded
- format, warnings-as-errors compile, strict Credo, erlfmt, native formatter/harness gates
- native Android and iOS simulator deploys; 20 repeated RPC reads on each returned the native scheme with availability status

Bead: app-7mu

### DIFF
--- a/lib/mob/theme.ex
+++ b/lib/mob/theme.ex
@@ -1,5 +1,5 @@
 defmodule Mob.Theme do
-  @color_scheme_nif_status {__MODULE__, :color_scheme_nif_status}
+  @nif_status {__MODULE__, :nif_status}
 
   @moduledoc """
   Design token system for Mob apps.
@@ -303,19 +303,11 @@ defmodule Mob.Theme do
 
     json = IO.iodata_to_binary(:json.encode(stringify_keys(payload)))
 
-    try do
-      :mob_nif.set_theme(json)
-    rescue
-      _ -> :ok
-    catch
-      _, _ -> :ok
-    end
+    nif_call(:ok, fn -> :mob_nif.set_theme(json) end)
   end
 
   defp safe_platform do
-    :mob_nif.platform()
-  rescue
-    _ in [UndefinedFunctionError, ErlangError] -> :host
+    nif_call(:host, fn -> :mob_nif.platform() end)
   end
 
   defp resolved_font_fallback(theme, platform) do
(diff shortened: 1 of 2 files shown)


GenericJam:
Addressed the adversarial review in 3fd3df1:

- platform, set_theme, and color_scheme now share one serialized VM-wide availability state
- unavailable callers skip native invocation until mob_nif is observably loaded; available-call failures transition back to unavailable
- isolated subprocess regression forces unknown state, asserts exactly one failed load across 12 concurrent mixed set/color callers, and explicitly cleans persistent state
- isolated fake-module lifecycle covers unavailable -> loaded -> available -> unloaded -> unavailable -> recovered
- full suite: 1282 passed, 38 excluded; all static/native harness gates green
- revised native Android and iOS deployments each passed 20 scheme reads plus Theme.set with shared status available

Ready for rereview.

GenericJam:
Adversarially reviewed. **Approve.**

Mechanism verified correct: lock-free persistent_term read in steady state, :global.trans-serialized first probe (12 concurrent callers → exactly one on_load report, proven by the new test), re-probe gated on :code.is_loaded flipping — and the device path can't regress because device NIFs are statically linked (ERL_NIF_INIT + -DSTATIC_ERLANG_NIF on both platforms), so first-load is deterministic and this PR only changes retry-after-failure behavior. Hot-code-push cycle (load→unload→reload) covered by the new recovery test.

Empirical: full-suite 'Failed to load NIF library' warnings 199 → 78 (Theme's own sites fully quiet — focused theme run 17 → 1, and that 1 is the by-design single probe report). Gates: 1282 tests / 0 failures, format, credo --strict, warnings-as-errors all clean. Rebase: branch sits directly on the 0.7.35 tag, no conflicts.

Non-blocking notes: (1) a genuine device NIF failure now logs once then falls back silently forever — an explicit one-shot Logger.warning in mark_nif_unavailable/1 would make that intentional rather than incidental; (2) recovery from :unavailable relies on another caller loading mob_nif — fine given the renderer, worth a comment; (3) the shared nif_call rescues broader than safe_platform/0's old UndefinedFunctionError/ErlangError set — a transient badarg briefly latches :unavailable (self-heals); (4) the remaining 78 warnings are Mob.Device/DNS probes outside this PR's scope — filing a follow-up issue.

Merging.

PR #120, opened 2026-09-03
MOB-134: apply throttle config to the table being built (iOS half)

GenericJam:
Every `Mob.UI` throttle/debounce setting was silently discarded on iOS. The call happened on every frame and always resolved to NULL.

`mob_set_throttle_config` is only reached from the `set_root` prop deserialiser, which runs while the frame is still being **built** — the swap that makes that table active is ~50 lines later. The handles in that JSON carry `tap_build_generation`, but `mob_resolve_active_tap_locked` compares against `tap_table_generations[tap_active]`, still the previous frame's generation. The comparison failed, the `if (tap)` body was skipped, and an app asking for `throttle: 100, delta: 8` quietly ran at the built-in 33ms/1.0 default.

Now resolves against the building table first, with an active-table fallback. Safe by ordering: `clear_taps` zeroes the building table's throttle fields at the top of the frame, `register_tap` writes only `pid`/`tag_env`/`tag`, and the swap loop copies only `identity_start_generation`.

**Android needs no equivalent change**, for a non-obvious reason: its swap happens inside `nif_set_root` *before* Kotlin composes, so by the time Kotlin could apply config the handles are already active. Android's gap is that nothing calls the function at all — the remaining half of this issue.

## Verification

iPhone 17 Pro simulator, two scroll nodes on one screen, one default-throttled and one `throttle: 500, delta: 1`, comparable one-second swipes:

| handler | events |
|---|---|
| default 33ms | **33** |
| configured 500ms | **5** |

Before the change both ran at the default. 1383 tests pass.

## The caveat matters more than the fix

The measurement only holds **if the handler does not re-render**.

`Mob.Screen.Server.forward/2` calls `paint/2` unconditionally after every `handle_info`, whether or not assigns changed. So a scroll handler on the screen process re-renders on every event; each render calls `clear_taps`, which zeroes `last_emit_ns`; and `mob_throttle_check` treats `last_emit_ns == 0` as "no previous emission" and emits unconditionally. **The throttle is defeated by the very events it is throttling.**

My first attempt at this measurement read 68 vs 69 and looked like the fix had failed. It had not — routing the same handlers to a plain process instead of the screen gives the 33-vs-5 above.

That is MOB-124-shaped (rendering driven by message arrival rather than state change, plus per-frame handle churn) and is recorded on the issue rather than worked around here.

No version bump — release hold still in effect.

### DIFF
--- a/android/jni/mob_nif.zig
+++ b/android/jni/mob_nif.zig
@@ -972,6 +972,12 @@ const TapHandle = extern struct {
     identity_start_generation: u32,
 
     // ── Batch 5 throttle state — populated by mob_set_throttle_config ──
+    // Set once the app has actually configured this slot. Without it, 0 is
+    // ambiguous: `throttle: 0` is a documented escape hatch meaning "raw firing
+    // rate", but a zeroed slot also means "never configured, use the default".
+    // Mob.Event.Throttle documents `throttle: 0` and has a doctest for it, so
+    // the one value a user reaches for was the one that could not be expressed.
+    throttle_configured: c_int,
     throttle_ms: c_int,
     debounce_ms: c_int,
     delta_threshold: f64,
@@ -1024,7 +1030,14 @@ fn tapGrowLocked(which: usize, needed: usize) bool {
     // Zero the new tail: clearTaps and the resolvers key off tag_env being null
     // to decide a slot is free, and realloc leaves it uninitialised.
     var i = tap_table_capacity[which];
-    while (i < cap) : (i += 1) table[i] = std.mem.zeroes(TapHandle);
+    while (i < cap) : (i += 1) {
+        table[i] = std.mem.zeroes(TapHandle);
+        // leading/trailing are the two fields whose "unset" value is 1, not 0.
+        // clearTaps resets them to 1 for reused slots, so without this a slot's
+        // default would depend on whether it arrived by growth or by reuse.
+        table[i].leading = 1;
+        table[i].trailing = 1;
+    }
 
     tap_tables[which] = table;
     tap_table_capacity[which] = cap;
@@ -1331,6 +1344,7 @@ pub export fn mob_set_throttle_config(
     erts.enif_mutex_lock(tap_mutex);
     defer erts.enif_mutex_unlock(tap_mutex);
     const h = resolveActiveTapLocked(handle) orelse return;
+    h.throttle_configured = 1;
     h.throttle_ms = throttle_ms;
     h.debounce_ms = debounce_ms;
     h.delta_threshold = delta_threshold;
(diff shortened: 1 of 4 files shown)


GenericJam:
**Adversarial review done; acted on it.** The review verified the mechanism — including the one thing I most wanted checked, that the build-table lookup cannot read past the build table's capacity. It cannot: `set_root` zeroes `tap_build_count` under the same lock as the table flip, and `register_tap` only increments after a successful grow, so `slot < tap_build_count` implies `slot < capacity`. The MOB-133 stale-count hazard does not reach this path.

It then found a second, independent reason config did not work — one that would have survived this fix and made it look ineffective.

## `throttle: 0` collided with "unset"

Both `mob_throttle_check` and the Zig `throttleCheck` did:

```c
int throttle_ms = h->throttle_ms ? h->throttle_ms : default_throttle_ms;
```

`Mob.Event.Throttle` documents `throttle: 0` as the raw escape hatch, **with a doctest**, and `renderer.ex` repeats it. So the one value an app reaches for to *disable* throttling was the one value indistinguishable from "never configured". The iOS struct comment already said `// 0 = no throttle (raw firing)` — the code contradicted its own stated intent. Same for `delta: 0`.

Fixed with an explicit `throttle_configured` flag on both platforms. Verified on the Pixel 8 emulator, three scroll nodes, one gesture each:

| list | events |
|---|---|
| A default 33ms | 20 |
| B configured 500ms | 4 |
| **C raw (`throttle: 0, delta: 0`)** | **41** |

C is the case that could not previously exist — before this it was indistinguishable from A.

## Correcting the scope claim

This makes **throttle and delta** work. `debounce_ms`, `leading` and `trailing` are carried on the wire, stored by `mob_set_throttle_config`, and **read by nothing** on either platform. So MOB-134's framing ("throttle/**debounce** config never reaches native") is only half fixed — debounce still does not work, because nothing implements it. That is now stated in the decision doc rather than left implied.

Also aligned the `leading`/`trailing` default between the two paths that set it: `clear_taps` resets them to 1 while a freshly grown slot was `memset` to 0, so a slot's default depended on whether it arrived by growth or reuse.

## Tests

The iOS half shipped none; it now adds four to the existing `native_event_handle_test.exs`, pinning the three invariants this fix rests on and that are invisible at a glance — `clear_taps` zeroes the **building** table, `register_tap` writes only routing fields, the swap loop copies only `identity_start_generation`. A reasonable-looking "zero the slot before writing it" tidy-up in `register_tap` would otherwise revert this silently. 1387 tests pass.

## Verification status, stated honestly

- **Android**: the full three-way result above is on-device.
- **iOS**: config-lands was verified on the simulator before this commit (instrumented `resolved=build, throttle=500`, and 33 vs 5 events), and events still arrive after it. **`throttle: 0` yielding raw firing is not iOS-device-verified** — the mechanism is symmetric with the Android one that is, and is pinned by the new source tests, but I am not claiming the measurement.

PR #121, opened 2026-09-03
MOB-140: skip the repaint when the rendered tree is unchanged

GenericJam:
`Mob.Screen.Server.forward/2` repainted unconditionally after every `handle_info`, with no comparison against the previous state. **Any** message to a screen triggered a full render — walk the tree, `clear_taps`, a `register_tap` per interactive node, serialise, `set_root`, native tree rebuild. A handler that changed nothing cost exactly as much as one that changed everything, so a 30 Hz scroll handler drove 30 full renders per second.

It also fed a feedback loop: each render calls `clear_taps`, which zeroes the per-handle throttle state including `last_emit_ns`, and `throttleCheck` reads that as "no previous emission" and emits unconditionally. **The native throttle was defeated by the very events it was throttling.**

MOB-134 measured it — the same two handlers, differing only in whether their delivery caused a repaint:

| handlers delivered to | default 33ms | configured 500ms |
|---|---|---|
| the screen process | 68 | 69 |
| a plain process | 33 | 5 |

## Compare the tree, not the assigns

Assign comparison is the LiveView model and would be cheaper, but it is wrong here. `render/1` may legitimately read state that is not in assigns — `Mob.Theme.set/1` is the obvious one, and a handler that changes the theme without assigning anything must still repaint; an assigns-based check would leave the old theme on screen. Running `render/1` and comparing its output keeps every such input honest, and still removes the expensive half, since the native crossing dominates per MOB-124's own measurements.

The tree is comparable across frames because handles are assigned later, inside `Mob.Renderer` — at the comparison point interactive nodes still carry their `{pid, tag}`, stable for the life of a screen.

Navigations, activation tokens and `:sync` callers always paint. Skipped repaints are recorded via `RenderStats.drop_frame/1` as uncommitted frames, so the meter reports how many were skipped rather than hiding them.

## Result

Pixel 8 emulator, both handlers on the screen and neither assigning — the shape that previously gave 68 vs 69:

| list | events |
|---|---|
| A default 33ms | 18 |
| B configured 500ms | **4** |

## Tests

Six, counting `set_root` at a stub NIF rather than reasoning about the code, covering both directions: a no-op message, re-assigning the same value, a real change, and a change `render/1` reads from outside assigns. Mutation-checked — restoring the unconditional repaint fails exactly the three "must not repaint" tests and leaves the three "must repaint" ones green.

One trap worth flagging for anyone extending them: settling with `Mob.Sender.sync/1` inside a `catch` reads the counter too early when the sync exits, which turns every "must repaint" assertion into a silent false failure. They wait for the counter to stop moving instead.

1393 tests pass. No version bump — release hold still in effect.

## Deliberately not covered

`render/1` and the expansion passes still run on every message; this removes only the native crossing. Whether a screen should run `render/1` at all for a message it ignored is a separate question. Per-frame handle churn is untouched and remains MOB-124's subject.

### DIFF
--- a/decisions/2026-09-03-skip-the-repaint-when-the-tree-is-unchanged.md
+++ b/decisions/2026-09-03-skip-the-repaint-when-the-tree-is-unchanged.md
@@ -107,3 +107,36 @@ different change, and the honest place to look next is whether a screen needs
 to run `render/1` at all for a message it ignored. It also does not address the
 per-frame handle churn itself — handles are still re-registered on every frame
 that does paint, which is MOB-124's subject.
+
+## What this does not do
+
+Recorded because each is a render input that lives outside the fingerprint, or
+a cost the headline numbers do not show.
+
+* **`render/1` and the expansion passes still run on every message.** Only the
+  native crossing is skipped. Whether a screen should run `render/1` at all for
+  a message it ignored is a separate question.
+
+* **Per-frame handle churn is untouched.** Handles are still re-registered on
+  every frame that does paint; that is MOB-124's subject.
+
+* **The listener pid is not in the key.** `Mob.Renderer.register_handler/2`
+  resolves tap targets through `Mob.Listener.handler/1`, a `Process.whereis`.
+  If the listener dies and restarts it gets a new pid, the committed tap table
+  still names the dead one, and every control on screen is inert. Previously
+  the next message repainted and re-registered, self-healing in one message; an
+  unchanged tree now never re-registers. Narrow — the listener is started once
+  and is not supervised for restart — but it is the same class as the theme bug
+  above, and closing the theme half does not close this one.
+
+* **A frame the sender rescued is still recorded as displayed.**
+  `Mob.Sender.commit/1` rescues a raising render in the sender process, after
+  the screen has already written `last_frame`. Nothing propagates back. Mostly
+  benign, since re-rendering the same tree would raise again, but the
+  observable change is that `[mob] render failed` now logs once instead of once
+  per message — a screen wedged by an unencodable prop goes quiet.
+
+* **Every painting frame pays an extra full tree traversal.** `phash2` over the
+  expanded tree runs whether or not the frame is skipped: roughly 70 ns/node
+  here, so ~15-150 µs on a 200-2000 node screen. A clear win when it skips,
+  pure added cost when it does not.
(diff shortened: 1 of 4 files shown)


GenericJam:
**Second review: both blockers verified fixed; merging after its three edits.**

The review enumerated every non-tree input `Mob.Renderer.render/4` reads — colours, spacing, radii, type scale, flags, fonts — and confirmed each is a pure function of the `%Mob.Theme{}` struct, so the theme genuinely closes B1. It also traced every repaint path (`handle_cast`/`handle_call`/`handle_info` clauses, all five router entry points, every `make_current/3` follow-up) and confirmed `repaint_if_changed/1` has exactly one caller, so B2 is airtight. It separately confirmed `last_frame` is never persisted, so a restarted screen cannot inherit a stale fingerprint.

Its three required edits, all done:

**A test that could not fail.** The fixture guard asserted `File.read!(__ENV__.file) =~ "text_color: :on_background"` — and its own explanatory comment contained that substring, so reverting the fixture to a resolved ARGB left it green. Now asserts the rendered tree, and mutation-checked by reintroducing exactly the original mistake.

**The collision claim was wrong.** I wrote that a collision costs one repaint and "the next actual change repaints normally". It doesn't: `last_frame` still holds the *old* tree's hash, so every later frame rendering the same new tree hashes identically and is skipped again. The screen stays wrong until the tree moves to a third value — on a settled screen, until the next user interaction. The probability and range were right; the failure mode was not. Corrected in both the comment and the ADR.

**Skips poisoned the meter they were meant to inform.** `drop_frame/1` already records superseded, inactive and navigation-discarded frames, and skips vastly outnumber them — at the ADR's own 30 Hz example the 500-entry ring is entirely no-op skips within ~17 seconds, evicting the committed frames and navigation-boundary drops MOB-124 is measuring. Skips now carry `reason: :unchanged`.

Also declared `:last_frame` in `Mob.Socket.t/0`, and the ADR gained a **"what this does not do"** section recording the things that are genuinely still open: `render/1` and expansion still run per message; handle churn is untouched; **the listener pid is a render input outside the fingerprint** (if `Mob.Listener` restarts, an unchanged tree never re-registers and every control goes inert — same class as the theme bug, and the theme fix does not close it); a sender-rescued frame is still recorded as displayed; and every painting frame pays an extra ~70 ns/node traversal the headline numbers don't show.

1396 tests pass under three seeds, credo clean, CI green.

PR #122, opened 2026-09-03
MOB-127: key children on the author's :id instead of on position (iOS)

GenericJam:
Nine `ForEach` sites in `MobRootView.swift` used `id: \.offset`. Positional identity means an insert or delete makes every later row a different view to SwiftUI, so it is rebuilt and its state discarded rather than moved.

Paired with mob_new#50 — one issue, two repos.

## The identity was on the wire but not reachable

The epic says "the identity already exists on the wire". True, but not where `ForEach` needed it: the `:id` prop was read into `MobNode` only as `nativeViewId`, and only for `native_view` nodes. `accessibilityId` is a different prop (`:accessibility_id`). So this adds `MobNode.nodeId`, populated from `MOB_PROP_id` for every node type, and `mobIdentifiedChildren/1` which pairs each child with a stable key. All seven `childNodes` sites use it.

## Three details that are easy to get wrong

- **Authored ids and positions are prefixed differently** (`i\x01…` vs `p\x01…`), so an author id of `"3"` cannot collide with position 3 and silently merge two unrelated rows.
- **A duplicate id folds in its position** rather than emitting two equal keys. SwiftUI misbehaves quietly on duplicate `ForEach` ids — and Compose *throws* on duplicates in a lazy list, so on the Android side this would turn a cosmetic problem into a crash.
- **An empty `id: ""` is treated as absent**, or every unnamed-but-present sibling would share a key.

## Verification

Behavioural proof is on the Android side (mob_new#50), since SwiftUI's diffing has no host-side observable: four rows each containing a `text_field`, plus a button that prepends a row. Typed `ZZZ` into "charlie", prepended — `ZZZ` stayed with charlie as it moved down. The same screen with the `:id` on the inner `text_field` instead of the repeated row keys positionally and loses the text, which is the before-behaviour on the same build.

iOS is source-asserted (6 tests) and confirmed to compile and render on the iPhone 17 Pro simulator.

Mutation-checked: reverting one site to positional fails two of the six.

## Deliberately not done

`MobTabView`'s `ForEach(Array(tabs.enumerated()), id: \.offset)` still keys positionally — it iterates tab dictionaries rather than child nodes, and a tab bar is short and rarely reordered.

`MobFrameTracker`'s `.onChange(of: id)` workaround is kept, and its comment corrected. It existed because trackers were not bound to one id for their lifetime; that is now true only for unnamed nodes, which still fall back to position.

**Worth knowing:** identity attaches to the repeated element, not to something inside it. An author who puts `:id` on a child of the repeated node gets positional keying and no warning — recorded in the decision doc as the most likely mis-use.

1402 tests pass. No version bump — release hold still in effect.

### DIFF
--- a/decisions/2026-09-03-children-key-on-author-id.md
+++ b/decisions/2026-09-03-children-key-on-author-id.md
@@ -67,9 +67,37 @@ author who puts `:id` on a child of the repeated node gets positional keying
 and no warning. That is worth documenting in the guides, and is the most
 likely way this is mis-used.
 
-**Not done here.** `MobTabView`'s `ForEach(Array(tabs.enumerated()), id: \.offset)`
-still keys positionally — it iterates tab dictionaries rather than child nodes,
-and a tab bar is a short, rarely reordered list. The `MobFrameTracker`
-`.onChange(of: id)` workaround is kept: named nodes no longer need it, but
-unnamed ones still fall back to position, so the hazard it guards is real for
-them.
+**`:id` is coerced to a String in `Mob.Renderer`.** iOS reads it as an
+`NSString` and ignores anything else; Android's `MobNodeIdentity` canonicalises
+any JSON value. So `id: user.id` with an integer — the most natural way to
+write it — keyed rows on Android and fell back to positional on iOS, silently.
+Coercing atoms and numbers once, before the value leaves Elixir, makes the two
+platforms agree by construction. Map and list ids still fall back to position on
+both.
+
+**`MobTabView` is keyed too.** An earlier version of this document claimed its
+`ForEach` "iterates tab dictionaries rather than child nodes" and left it
+positional. That was wrong — the body subscripts `node.childNodes`, so toggling
+a conditional tab shifted every later tab's entire content subtree into its
+neighbour's identity.
+
+**`MobFrameTracker`'s `.onChange(of: id)` is kept, for a different reason than
+first written.** A tracker only exists when the node has an id, so unnamed nodes
+have no tracker and the original "unnamed nodes still shift" rationale was
+wrong in both directions. What remains is the duplicate-id fallback: those keys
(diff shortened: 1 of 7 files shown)


GenericJam:
Second adversarial review returned do-not-merge on two blocking findings; both are fixed in `b9a901e` (this repo) and `1b3dc96` (mob_new#50).

**Blocking — Android covered 3 of 7 containers.** column, row and the lazy list were keyed while box, both scroll axes and the sheet body were still positional, so "keyed on both platforms" was false and MOB-127 stayed unfixed for four containers as a *new* divergence. Fixed in mob_new#50; re-verified on the Pixel 8.

**Blocking — the decision record described deleted code.** It credited `MobNode.nodeId` and `MobNodeIdentity.keyFor`, both removed by the second commit, and its Context claimed `:id` was only read for native_view nodes — `git show origin/master:ios/mob_nif.m` shows that read has always been unconditional. Rewritten, duplicate paragraph dropped, and the coercion's real scope stated: prepare_props only, so `tabs: [%{id: 1}]` still degrades to positional.

**Narrowed the id coercion to numbers.** The atom half changed nothing — `:json.encode` already stringifies bare atoms — while turning `true`/`false` from a JSON boolean both platforms reject into the real id `"true"`. Verified by running the encoder rather than reasoning about it.

**MobTabView now has tests.** It was fixed only because the first review caught a false claim of mine, and reverting the fix passed every test on the branch. Two tests cover it now.

Also filed, not fixed here: MOB-141. Author-id keying means replacing a list's contents re-fires `.onAppear` on the last row, so a search screen re-queried per keystroke now fires one `on_end_reached` per keystroke. Android's index-based `derivedStateOf` has always done this, so this converged iOS onto it rather than inventing a third behaviour — which is why the fix belongs in one place for both platforms, not in this PR.

1405 tests pass. `mix.exs` untouched at 0.7.38.

PR #123, opened 2026-09-03
Record the rendering-performance epic in the changelog

GenericJam:
Docs only. Five merged issues since 0.7.38 had no changelog entry: MOB-124 (`Mob.RenderStats`), MOB-133 (tap tables grow on demand), MOB-134 (throttle config targets the building table), MOB-140 (skip the unchanged repaint) and MOB-127 (children key on the author's `:id`).

Written now rather than at release, so each entry is drafted while the reasoning is still at hand instead of reconstructed from commit subjects weeks later.

No version bump — the epic is not finished, and the standing constraint is to release nothing until it is.


GenericJam:
Branched from a local checkout whose history has diverged from origin/master, so the diff swept in the entire tree instead of one file. Redoing from a worktree that tracks master.

PR #125, opened 2026-09-03
Cache decoded images instead of re-reading them from disk every render

GenericJam:
Found while investigating MOB-126.

`UIImage(contentsOfFile:)` opens, reads and decodes the file synchronously, and unlike `UIImage(named:)` there is **no system cache behind it**. It was being called from inside `MobImage.body`, so that read and decode ran on the main thread on every evaluation of the node. There was no image cache of any kind anywhere in `ios/`.

Bad in the steady state, much worse across a navigation: the root carries `.id(currentNavVersion)`, so a push or pop rebuilds the whole view tree and every local-file image on the incoming screen is decoded from scratch **while the transition is mid-animation**. That is one of the things actually behind the symptom MOB-126 was filed for.

### What it does not fix

A first visit still decodes cold. What goes away is every decode after that: re-rendering a screen already up, and navigating back to one visited before.

### Design

- **Key is (path, size, mtime)**, not the path alone. A file can be rewritten in place, and path-only keying serves the superseded image for the rest of the process, which is a correctness bug rather than a performance one. Costs one `stat(2)` per body evaluation, knowingly left on the main thread, buying "never stale" for about a thousandth of a decode.
- **Cost is decoded bytes** off the backing `CGImage`, not file size. A 2 MB JPEG is roughly 14 MB of bitmap, so a file-size budget would let the cache hold several times what it believes it holds.
- **Budget scales with the device**: a sixty-fourth of physical memory clamped to 16-64 MB, rather than a flat 64 MB. A flat number is only defensible on the largest device it runs on, and Mob targets old hardware deliberately. `NSCache`'s memory-pressure eviction is a backstop, not the plan.
- **No negative caching**, so a path that becomes valid later is not pinned to the placeholder for the process lifetime. The miss path is unchanged.
- The `AsyncImage` http(s) branch is untouched; URLSession already caches those.

### Tests

12 source-contract tests, **each mutation-tested against a broken implementation** (cache removed, path-only key, cost as file size, placeholder dropped, plain Dictionary instead of NSCache, negative caching added, no cost limit, http branch rerouted). All were caught.

They assert against comment-stripped source via a small Swift-aware scanner that copies string literals verbatim, so the `"http://"` prefix check is not eaten as a comment. That matters here: one mutation reverts the code while leaving a comment naming every symbol the tests match on, and without stripping several assertions pass against code with the bug back.

1417 tests pass. `mix.exs` untouched at 0.7.38.

### DIFF
--- a/ios/MobRootView.swift
+++ b/ios/MobRootView.swift
@@ -1780,8 +1780,11 @@ private struct MobSheetView: View {
 }
 
 // ── Local-file image cache ───────────────────────────────────────────────────
-// UIImage(contentsOfFile:) opens the file, reads it and decodes it in full,
-// synchronously, and unlike UIImage(named:) there is no system cache behind it.
+// UIImage(contentsOfFile:) reads the file and builds a CGImageSource-backed
+// image; the decompression itself is deferred to first draw, which is why
+// UIImage.preparingForDisplay() exists. Either way it is synchronous work on
+// whichever thread touches it, and unlike UIImage(named:) there is no system
+// cache behind it, so nothing survives to the next render.
 // The asset-catalogue path gets one for free; the loose-file path gets nothing.
 // Calling it from a SwiftUI `body`, as MobImage used to, means that
 // read-and-decode runs on the main thread on every evaluation of the node. That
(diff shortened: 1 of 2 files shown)


GenericJam:
## Adversarial review: DO-NOT-MERGE, one blocking. Fixed.

**B1 — the tests did not test the cache.** They asserted the write side and the call site and said nothing about the read. The reviewer ran three mutations that completely defeat the feature; all passed 12/12:

| Mutation | Effect |
|---|---|
| delete the `cache.object(forKey: key)` lookup | can never hit; decodes every render, plus a stat. Worse than master |
| `static let shared` → a computed property | fresh empty cache every access, 0% hit rate for ever |
| read under a constant key | serves one image for **every path in the app** |

My commit message claimed each test was mutation-tested against a broken implementation. That was not true of the three most obvious breakages. All are caught now, along with a fourth the reviewer found: the budget divisor could go 64 → 4, a 16x blowout on exactly the low-memory hardware the scaling exists to protect, while staying inside the clamp the test asserted.

### Also fixed

- **NSCache silently DROPS an object whose cost exceeds `totalCostLimit`** — the insert appears to succeed and the next lookup returns nil (verified against Foundation). A 12 MP capture is ~48.75 MB decoded, over the budget of every device under 4 GB. So the most likely large image in a real app was re-decoded every render exactly as before, plus a stat, while the code read as though it were cached — and **scaling the budget down for small devices made this bite hardest on the ones it was meant to help.** Oversized images now skip the insert. That is the honest floor, not a fix; the fix is downsampling at load, which changes what is rendered rather than what is remembered.
- **`attributesOfItem` does not traverse a terminal symlink** while `UIImage(contentsOfFile:)` does, so a symlinked path was keyed on the link's size and mtime and never noticed the target being rewritten. Reviewer demonstrated it end to end. That is precisely the staleness the key exists to prevent, defeated in the one case where a path is most likely to be a stable alias for changing content.
- **A rewritten path leaked its old entry**, holding budget until NSCache evicted it — and what NSCache evicts to make room is some other screen's images.
- **A missing file attribute defaulted to 0**, degrading the key to path-only: the exact bug the function exists to avoid, against the policy stated four lines above it.
- Corrected the header comment: `UIImage(contentsOfFile:)` does **not** decode eagerly; it builds a CGImageSource-backed image and decompression is deferred to first draw. The cache still works, but not for the reason the comment gave. The forward-navigation claim is narrowed to match.
- Stripper fixes: nested block comments, and raw strings (`#"a\\"#` desynchronised the scanner for the rest of the file). Both fail in the under-stripping direction, which is the one that makes an assertion satisfiable by a comment.

18 tests, 8 mutations run and all caught. 1423 tests pass.

PR #126, opened 2026-09-03
MOB-126: measure the native half of a frame

GenericJam:
MOB-126's premise did not survive the source; the full investigation is on the Linear issue. Short version: `.id(currentNavVersion)` is the **only** thing that makes the transition fire (SwiftUI decides insertion/removal by identity, not by value), Android does the same thing via `AnimatedContent(contentKey:)`, and the JSON to node rebuild it is blamed for happens unconditionally anyway. Options 1 and 3 in the issue are not implementable; option 2 is the substance of MOB-129.

What is real is that **nobody has ever measured the native half of a frame.**

`Mob.RenderStats` measures seven stages and every one is on the BEAM side. `set_root_us` closes when `nif_set_root` returns, and `nif_set_root` finishes by calling `setRoot:transition:`, which dispatches to the main thread and returns. Everything SwiftUI then does to build, lay out and display the tree happens after that measurement closed. Compose is the same shape.

That did not matter until now. MOB-129 is a large change justified entirely by the cost of the navigation teardown, and MOB-130 says the wire-format decision must be made *on evidence only*. Neither can be settled against a number nobody has.

### What is measured

Main-thread busy time, from the tree being applied to the run loop going idle. Deliberately the pessimistic reading (anything else on the main thread in that window counts) and also the honest one: a frame drops when the main thread is busy, whoever made it busy. Documented as an upper bound, not an attribution.

**The closing bracket is a `beforeWaiting` run loop observer.** `CATransaction.setCompletionBlock` is the obvious alternative and is wrong: SwiftUI's update routinely lands in a later transaction than the one open at assignment time, so the completion fires *before* the measured work happens and the number reads near zero. A test pins this, because the wrong version fails silently by reporting something plausible.

**The transition travels with each sample** and `native_summary/1` groups on it rather than pooling. A `"none"` sample re-renders into an existing tree; `"push"`/`"pop"`/`"reset"` rebuild it. The size of that gap is exactly what MOB-126 and MOB-129 argue about.

**Off by default; the disabled path is one relaxed atomic load.** `_Atomic` rather than lock-guarded, so an app that never measures does not take, on the main thread, on every `set_root`, a lock a NIF thread also takes.

Reads degrade to `{:error, :unsupported}` rather than raising: the host has no `:mob_nif`, and a function exported from `mob_nif.erl` but absent from the native library keeps its Erlang stub and raises for ever.

### Scope

**iOS only.** The Android half needs a Choreographer or `doOnPreDraw` bracket in `MobBridge.kt.eex`, which MOB-142 is editing in the same window; sequencing avoids a conflict in a template that has already been the site of two cross-platform divergences in this epic. Android reports `:unsupported`, which is accurate rather than silently zero.

### Tests

16, each mutation-tested: gate removed, `beforeWaiting` swapped for a CATransaction completion, transition dropped, encode moved inside the lock, enable no longer clearing the window, `_Atomic` downgraded to plain `int`, NIFs absent from the function table, and exported-but-not-in-`-nifs` (a bug this repo has shipped: `load_nif` succeeds and the function raises for ever). All 8 native and 3 Elixir mutations were caught.

1421 tests pass. `mix.exs` untouched at 0.7.38. Decision record: `decisions/2026-09-03-measure-the-native-half-of-a-frame.md`.

Native iOS build verified against a real app (mob_plugin_demo pointed at this worktree): the NIF compiles and links. Runtime verification over dist to follow in a comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,24 @@ Everything below is unreleased work from the MOB-124 rendering-performance
 epic. Nothing here has shipped to Hex.
 
 ### Added
+- **Native frame timing: `Mob.RenderStats.native_enable/1`, `native_frames/1`
+  and `native_summary/1`** (MOB-126). `Mob.RenderStats` measures seven stages
+  and every one is on the BEAM side of the boundary: `set_root_us` closes when
+  `nif_set_root` returns, and that is the moment the tree is handed to the main
+  thread, not the moment it is on screen. Everything SwiftUI does to build, lay
+  out and display it happened after the measurement closed, so the native half
+  of every frame had never been measured. These read a native ring buffer of
+  main-thread busy time per applied tree, tagged with the transition that
+  produced it so a navigation rebuild can be told apart from a steady-state
+  re-render.
+
+  Off by default; the disabled path is one relaxed atomic load per `set_root`.
+  Read it as an upper bound on a frame's native cost rather than an
+  attribution: anything else queued on the main thread in the same window is
+  inside the number. **iOS only for now** — Android returns
+  `{:error, :unsupported}`, which is accurate rather than silently zero. See
+  `decisions/2026-09-03-measure-the-native-half-of-a-frame.md`.
+
 - **`Mob.RenderStats` — per-frame render instrumentation** (MOB-125). Records
   the user's `render/1`, tree expansion, component reconcile, the renderer's
   prepare walk, `register_tap`, `:json.encode`, and `set_root` as seen from the
@@ -109,6 +127,19 @@ epic. Nothing here has shipped to Hex.
   recorded through `Mob.RenderStats` with `reason: :unchanged` rather than
   vanishing, so a screen that stops updating stays diagnosable. See
   `decisions/2026-09-03-skip-the-repaint-when-the-tree-is-unchanged.md`.
+- **Local-file images are cached instead of re-read on every render**
+  (found via MOB-126). `UIImage(contentsOfFile:)` was called inside SwiftUI's
+  `body`, and unlike `UIImage(named:)` there is no system cache behind it, so
+  every evaluation of an image node re-read the file from disk on the main
+  thread. Worst across a navigation, where the root's identity change rebuilds
+  the whole tree and every image on the incoming screen is loaded again while
+  the transition animates.
+
+  Keyed on (path, size, mtime) so a file rewritten in place is not served
+  stale, with the budget scaled to physical memory (16-64 MB) rather than a
+  flat ceiling that is only defensible on the largest device it runs on. A
+  first visit still loads cold; what goes away is every load after that.
+
 - **Children keep their identity across list edits** (MOB-127). Children of
   every container were keyed on their position, so inserting or removing a row
   made each following row adopt the previous occupant's view state: typed text,
--- a/decisions/2026-09-03-measure-the-native-half-of-a-frame.md
+++ b/decisions/2026-09-03-measure-the-native-half-of-a-frame.md
@@ -1,0 +1,97 @@
+# Measure the native half of a frame
+
+**Date:** 2026-09-03
+**Issue:** MOB-126 (and the evidence gate on MOB-129 / MOB-130)
+
+## Context
+
+`Mob.RenderStats` was built for MOB-125 so the rendering-performance epic would
+stop being a set of guesses. It measures seven stages: the user's `render/1`,
+expansion, reconcile, the prepare walk, `register_tap`, `:json.encode` and
+`set_root`.
+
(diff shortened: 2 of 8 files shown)


GenericJam:
## Runtime verification on the iOS simulator

Built `mob_plugin_demo` against this worktree (`mob_dir` repointed) and deployed to an iPhone 17 Pro simulator. **iOS native build completes**, so the new ObjC compiles and links into a real app.

Over dist:

```
enable:  :ok
frames:  {:ok, %{"enabled" => 1, "recorded" => 1, "dropped" => 0,
                 "samples" => [%{"apply_us" => 81224.9998562038,
                                 "seq" => 0, "transition" => "none"}]}}
```

That exercises the whole path: the enable NIF flips the atomic and `enabled => 1` reads back from native, a real `set_root` arms the observer, the observer fires on run loop idle and writes a sample, and the reader NIF's JSON decodes through `native_frames/1`. 81 ms is the first cold build of a screen's tree on the simulator, which is the right order of magnitude for a first-ever SwiftUI build including font resolution.

## What is not verified

**No push/pop samples.** `mob_plugin_demo` never reaches its root screen on this simulator: its boot dies at step 5 on an Ecto/SQLite migration (`Exqlite.Sqlite3NIF.open/2` fails, the `DBConnection` checkout is killed). Unrelated to this change, but it means `:mob_screen` is never registered, so `Mob.Test.navigate` returns `:ok` as a no-op cast and no navigation `set_root` ever happens. The one sample above came from starting a root screen manually.

So the mechanism is proven and the transition **split** is unit-tested and mutation-tested, but the *comparison the split exists for* — how much more a `push` costs than a `none` on real hardware — is not collected here. That measurement is MOB-129's opening move, and it is the reason this lands first.

I did not chase the demo app's migration failure: it is an unrelated app-environment problem, and fixing someone else's app to get a nicer number in this PR is not the right trade.

GenericJam:
## Adversarial review: DO-NOT-MERGE, two blocking. Both fixed.

**B1 — every iOS release build would have failed to link.** `mob_native_stats_enabled` and `mob_record_native_frame` were defined inside the `#if !MOB_RELEASE` harness block while `MobViewModel` calls both on every `set_root` with no knowledge of that flag. The release pipeline compiles `ios/*.swift` without it and `mob_nif.m` with it, so the symbols are simply absent, and `-O -wmo` cannot strip the calls because they are live in `setRoot`. Debug links fine, which is exactly why my device verification did not catch it. The file already states the rule where `g_ui_event_seq` is declared: *only the harness reads it; the writers stay unconditional.* Storage and the two Swift-callable entry points now sit outside the guard, only the two reader NIFs inside, pinned by a test.

**B2 — the observer ran before the work it measured.** `beforeWaiting` observers fire in ascending `order`, and Core Animation's commit observer sits at 2000000. SwiftUI evaluates bodies and lays out *inside* that commit, so `order: 0` closed the bracket ahead of the entire apply and still reported a plausible number. That is the same class of silent-wrong-answer as the `CATransaction` completion this module explicitly rejects, and I walked into a different version of it. Now `CFIndex.max`.

**I could not settle B2 on device, and I am not going to pretend otherwise.** Single cold-start samples: **81 ms** before, **11 ms** and **5.5 ms** after. That is the opposite direction from the prediction — but two runs of the identical build differ by 2x, and the demo app dies at boot before producing a second sample, so nothing here separates the order effect from run-to-run variance. The change stands on the platform argument, not on those numbers, and **the earlier 81 ms figure in this PR should not be quoted as evidence of anything.**

### Also fixed

- **The lock test did not test the lock.** It located the critical section's close by the text that followed it, so hoisting the memcpy *above* `@synchronized` — the exact regression it is named for — still passed, while an added comment broke it. Now walks brace depth.
- **`decode_native/1` accepted any payload with a `samples` key**, so a scalar reached `length/1` and a list of bare numbers reached `Access.get/3`, both raising out of a function documented never to do that.
- **A non-numeric `apply_us` sorted above every number** and became p95 and max, capturing exactly the half of the distribution this exists to look at.
- **`rescue ErlangError` swallowed everything**, so a real `:system_limit` from a working NIF read as "this platform does not support it". Narrowing it surfaced a bug in the narrowing, caught by the new test: that clause also catches errors Elixir normalises into their own structs, and `SystemLimitError` has no `:original`, so reading it raised `KeyError` from inside a rescue.
- Two upward biases now documented: `beforeWaiting` fires only when the loop actually sleeps, and `.commonModes` misses run loop excursions outside the common set. Both poison `max` and `p95`.
- CHANGELOG entry added (it had none).

Rebuilt and redeployed to the simulator after both fixes: **iOS native build completes**, and the NIF still works end to end over dist. 1427 tests pass.

PR #127, opened 2026-09-04
MOB-141: fire on_end_reached on arrival at the end, not on replacement

GenericJam:
**The one thing in this unreleased window that is worse than before.** MOB-127 keyed children on the author's `:id`, so replacing a list's contents gives every row a new identity and the last row's `.onAppear` runs again even though nobody scrolled. A search screen re-queried on each keystroke fires one pagination request per keystroke, where before it fired none.

### The fix

Latched on the child count. That works because **only navigation changes the list's identity**, so the `@State` outlives a new tree arriving for the same screen. Re-querying and getting twenty results again is suppressed; loading a page and going twenty to forty is not, which is the pagination flow the callback exists for.

The latch is set **before** the callback, so a handler that synchronously triggers a re-render cannot re-enter and fire twice for the same count.

### Not a complete fix

Stated in the code rather than implied away: a re-query whose result count differs every time still fires once per distinct count. Separating "new content, user is at the end" from "new content, user never scrolled" needs **scroll position**, not content identity, and the honest version drives this from the scroll observer rather than `.onAppear`. Handlers should still be idempotent.

### Structure

Extracted into `MobLazyList` rather than adding `@State` to `MobNodeView`, which renders every node in the tree — an `Optional<Int>` there is storage paid thousands of times over for something one node type uses.

### Android is deliberately unchanged

Its `derivedStateOf` over `layoutInfo` has always been index-based and has always had this shape. So this repairs an **iOS regression** rather than making a cross-platform behaviour change. Converging the two is worth doing but it is a decision for Android users, not a regression fix.

### Also

Fixes a brittle assertion this refactor exposed: MOB-127's test counted the inline `ForEach(mobIdentifiedChildren(...))` spelling and broke when the call was hoisted into a `let`, though the identity property it guards was untouched. It now counts the derivation wherever it appears.

4 tests, each mutation-tested (latch guard removed, latch set after the callback, reverted to the unlatched inline form, `@State` back on every node). All caught. 1449 tests pass. `mix.exs` untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,22 @@ epic. Nothing here has shipped to Hex.
   flat ceiling that is only defensible on the largest device it runs on. A
   first visit still loads cold; what goes away is every load after that.
 
+- **`on_end_reached` fires on arrival at the end, not on content replacement**
+  (MOB-141). Keying children on the author's `:id` gave every row a new identity
+  when a list's contents were replaced, so the last row's `onAppear` ran again
+  even though nobody scrolled: a search screen re-queried on each keystroke
+  fired one pagination request per keystroke, where before it fired none. The
+  callback is now latched on the child count, which survives a replacement
+  because only navigation changes the list's identity. Re-querying and getting
+  twenty results again is suppressed; loading a page and going twenty to forty
+  is not.
+
+  Not a complete fix, and the limitation is worth knowing: a re-query whose
+  result count differs every time still fires once per distinct count.
+  Separating "new content, user is at the end" from "new content, user never
+  scrolled" needs scroll position rather than content identity. Write
+  `on_end_reached` handlers to be idempotent.
+
 - **Children keep their identity across list edits** (MOB-127). Children of
   every container were keyed on their position, so inserting or removing a row
   made each following row adopt the previous occupant's view state: typed text,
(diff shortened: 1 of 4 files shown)


GenericJam:
## Device-verified on the iOS simulator, both directions

The demo app had no screen using `lazy_list`, so the refactored path was covered by the suite and by source assertions but had never actually rendered. I added a probe screen that renders one, counts `on_end_reached` fires, and can either replace its contents at a constant count or grow them.

| step | rows | fires |
|---|---|---|
| mount | 8 | 1 |
| replacement 1 (same count, different content) | 8 | **1** |
| replacement 2 | 8 | **1** |
| replacement 3 | 8 | **1** |
| grow | 16 | **2** |
| grow | 24 | **3** |
| grow | 32 | 3 |

**The regression is fixed**: three wholesale content replacements, every row getting a new identity each time, and the callback does not re-fire. Before this change each one would have incremented.

**Pagination still works**: growing the list re-arms the latch and fires again, which was the thing most at risk from a latch that never released.

The last line is the most reassuring. At 32 rows the list exceeds the viewport, so the final row is below the fold and the callback correctly stays silent — it is genuinely tied to the last row *appearing*, not to the count merely changing. A latch keyed on count alone with no `.onAppear` gate would have fired there.

Also confirms the `MobLazyList` extraction renders correctly, which the suite could not tell me.

PR #128, opened 2026-09-04
Release prep 0.7.39: state the upgrade hazards the notes were understating

GenericJam:
Pre-publish review found **nothing wrong with the code** but three places where the notes describe behaviour a user will not actually get. Docs only.

**`throttle: 0` is a live upgrade hazard and only the safe direction was written down.** The MOB-134 entry framed the fix as previously-unthrottled handlers becoming throttled. The reverse also happens and is the dangerous one: `throttle: 0` is the documented escape hatch for raw delivery, and because the config never reached native, an app asking for it was silently getting the 33 ms default. It now means what it says, so a handler that has been receiving ~30 Hz starts receiving **every sample** — 120 Hz on ProMotion, into a screen mailbox.

**The native timing functions are unavailable in an iOS release build** — exactly when someone reaches for a profiler. The reading NIFs are inside the same `MOB_RELEASE` guard as the rest of the harness, so a TestFlight build returns `{:error, :unsupported}` indistinguishably from Android and from the host. Neither the `@doc` nor the changelog said so.

**"Both platforms derive keys identically" was false for the tab bar.** iOS keys it via `mobIdentifiedTabs`; Compose's `NavigationBar` still iterates positionally, so reordering a tab moves per-tab state on Android and not on iOS. The other seven sites do match.

Also: MOB-128 lacked the "needs mob_new 0.4.31+" note its siblings carry; the `on_end_reached` entry described one limitation of the count latch when there are three (a windowed list at constant length fires once ever; a failed page load can never be retried by scrolling); `debounce`/`leading`/`trailing` are stored but acted on by neither platform though the headline said "throttle/debounce reaches native"; `native_disable/1` was missing from the added list.

No version bump here — the bump is a separate commit so the release workflow fires on a reviewed tree. 1449 tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,8 @@ Everything below is unreleased work from the MOB-124 rendering-performance
 epic. Nothing here has shipped to Hex.
 
 ### Added
-- **Native frame timing: `Mob.RenderStats.native_enable/1`, `native_frames/1`
-  and `native_summary/1`** (MOB-126). `Mob.RenderStats` measures seven stages
+- **Native frame timing: `Mob.RenderStats.native_enable/1`, `native_disable/1`,
+  `native_frames/1` and `native_summary/1`** (MOB-126). `Mob.RenderStats` measures seven stages
   and every one is on the BEAM side of the boundary: `set_root_us` closes when
   `nif_set_root` returns, and that is the moment the tree is handed to the main
   thread, not the moment it is on screen. Everything SwiftUI does to build, lay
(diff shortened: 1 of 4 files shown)


GenericJam:
## Added: guide documentation for the three undocumented features

Prompted by a direct question about whether the new features are documented anywhere a user would actually look. They were not.

RELEASE.md requires new functionality to ship with docs "in the right place (module/function `@doc`, guides under `guides/`)". The `@doc`s existed; the guides did not, so all three were reachable only from the CHANGELOG.

| feature | before | now |
|---|---|---|
| `lazy: true` on `:scroll` | one passing mention at `components.md:802`, inside the handle-limits paragraph; absent from the `:scroll` prop table | in the table, plus a subsection with the measured win and the four silent consequences |
| `:id` child keying | **zero** across components, events, navigation, liveview | new `## Giving children a stable :id` section |
| `Mob.RenderStats` | **zero** | new troubleshooting section covering both halves |

The consequences are the part that mattered most to write down, because every one of them is silent: with `lazy: true`, rows below the fold register no frame so `element_frames` and `tap_id` cannot address them, `scroll_to(:bottom)` under-scrolls, `screenshot_tour` truncates, and scroll position becomes index-based. That is *why* it is opt-in, and it was documented nowhere a user would look.

For `RenderStats` the same applies to two traps: `register_tap_us` is nested inside `prepare_us` so adding them double-counts, and `apply_us` is an upper bound rather than an attribution.

`on_end_reached` also gained the semantics it never had — the row-count latch, why it exists now that children key on `:id`, and the three cases it does not cover.

**Deliberately not in this PR:** `MOB_FONTS.md` is referenced from `guides/styling.md`, `MOB_STYLES.md`, `MOB_PLUGINS.md` and linked from the CHANGELOG, but is absent from ex_doc's `extras`, so `mix docs` warns and HexDocs ships a broken link. The one-line fix touches `mix.exs`, which is the release trigger — merging it here would fire the workflow at the already-published 0.7.38. It goes in the version-bump commit instead, so one `mix.exs` change means one release.

1449 tests pass. `mix docs` builds with no warnings once that fix lands.

PR #132, opened 2026-09-04
Fix what MOB-129 broke by keeping parked screens alive

GenericJam:
MOB-129 merged **without** an adversarial review — the only substantial change in that batch that skipped one, and the largest. The review run after the fact (MOB-147) found two blocking defects. The process gap is closed separately in mob#131 / mob_dev#61 / mob_new#55.

### B1 — the frame registry refused every write from a screen you popped back to

Exactly the case MOB-129 optimises for. `mob_bump_frame_generation` still fires on every navigation and `mob_register_frame` still refuses a stale stamp — but a parked tracker is never removed from the hierarchy, so its `.onAppear` never fires again. A → push B → pop A left A's trackers holding generation 1 against a current 3, and **every write was rejected**. `Mob.Test.element_frames` and `tap_id` read empty for the screen actually on display, silently.

**The decision record asserted the opposite.** It reasoned about the outgoing direction — where the behaviour is correct and wanted — and never considered the returning one. That claim is corrected in place rather than quietly dropped.

Fixed by re-seeding the stamp when a slot becomes active, **not** by removing the gate: the gate is what stops an outgoing screen re-registering at mid-animation coordinates as it slides away, and what keeps two screens sharing an `:id` from clobbering each other.

**Verified on device:**

| step | screen | `element_frames` |
|---|---|---|
| A | dense | 231 |
| push B | home | 1 |
| pop back to A | dense | **231** |

Without the fix that last row is 0.

### B2 — toggles and sliders inherited state across screens

Neither had **any** prop-to-state sync; both relied on `.id(currentNavVersion)` destroying their identity to re-seed. With identities alive across navigation, a screen shows the toggle states and slider positions of the screen **two navigations back**.

The decisive evidence is in this repo: `lib/mob/socket.ex` raises `ArgumentError` on `transition: :none` because *"a TextField at the same position inherits the old screen's text and focus"* — and MOB-129 made that the default path for every push and pop.

### The rest

One new environment value, `mobScreenIsActive`, published per slot:

- a **parked sheet** is dismissed and re-presented on return. `.sheet` presents on the *window*, so the slot's `allowsHitTesting(false)` never reached it — it stayed visible **and interactive** over the incoming screen. A park is not reported to the BEAM as a user dismissal.
- a **parked video** pauses, audio included, and only resumes if it was autoplaying
- a **parked GPU view** stops rendering, rather than running at 60 fps indefinitely behind the visible screen
- **`on_end_reached`'s latch** clears on activation, restoring what the navigation teardown used to give it for free

### Tests

12 new, plus 5 added to the MOB-129 file for mutations the review **proved survived** — including inserting an unconditional `slots[outgoing] = nil`, which destroys depth-1 retention (the headline benefit) with all 13 original tests still green. **13 mutations run, all caught.**

The MOB-129 stripper is rewritten. The review demonstrated two holes: it stripped only full-line `//` comments, so deleting the real `containerWidth = width` and leaving the text in a trailing comment passed; and its dotall `/* */` strip swallowed ~200 lines of real code when a string literal containing `/*` was added.

### One review finding refuted by experiment

It predicted that seating the offset outside `withAnimation` coalesces away, so a push following a push would slide in from the wrong side. I recorded two consecutive pushes and extracted frames — **both enter correctly from the trailing edge.** The reviewer flagged that one as reasoning rather than verification and asked for exactly this experiment.

### Not fixed — recorded on MOB-147

A documented `reset_to(..., transition: :push)` retains an unreachable screen, because the clear keys on the animation name rather than on whether the stack was replaced; that needs a wire signal. A reset no longer cross-fades. And the layout cost of a parked slot on a container geometry change is reasoned about but unmeasured.

1479 tests pass. `mix.exs` untouched at 0.7.39.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,11 +43,30 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
   visible screen. `on_end_reached`'s latch clears on activation, restoring the
   pre-MOB-129 behaviour that a navigation used to give it for free.
 
-  Still open, and recorded on MOB-147: a documented `reset_to(...,
-  transition: :push)` retains an unreachable screen because the clear keys on
-  the animation name rather than on whether the stack was replaced; a reset no
-  longer cross-fades; and the layout cost of a parked slot on rotation is
-  reasoned about but unmeasured.
+- **A navigation that replaces the stack releases the screen it replaced**
+  (MOB-147). The release used to key on the animation name, which got it wrong
+  in both directions: `reset_to(..., transition: :push)` is documented, does
+  replace the stack, and was read as a push and retained for ever; while
+  `switch_tab(..., transition: :reset)` is also documented, does NOT replace the
+  stack — a parked tab can be switched back to — and was released, throwing away
+  the retention it should have kept.
+
+  The router now tags a stack-replacing navigation, and native reads the
+  animation from the prefix and the release from the suffix, so neither is
+  inferred from the other. No wire schema or NIF arity change: the transition
+  already travels as an atom whose name native reads directly, and the public
+  `:push | :pop | :reset` vocabulary callers use is unchanged.
+
+- **A reset cross-fades again** (MOB-147). The outgoing screen was released
+  synchronously, removing it in the same turn, so it hard-cut rather than fading
+  — and with `.transition()` gone there was nothing to animate its removal. The
+  release now runs on the animation's completion.
+
+  Still open on MOB-147: the layout cost of a parked slot on a container
+  geometry change. Steady state with both slots warm measures 67.6 ms against
+  65.7 ms for one, about 3%, which suggests the equality check is doing its job.
+  Rotation itself remains unmeasured, and honestly so: a geometry change
+  produces no `set_root`, so the frame instrumentation cannot see it at all.
 
 
 ### Performance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,15 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
   `transition: :none` citing this exact hazard; MOB-129 had made it the default
   path. Both now re-seed when the BEAM's value changes.
 
+  Toggles, sliders and text fields re-seed when their screen becomes active,
+  not only when the BEAM's value changes. A value watcher alone misses the
+  common case: slots alternate, so a screen reuses the view identities of the
+  screen two navigations back, and if both carry the same value — `false` for a
+  toggle, `""` for an uncontrolled field — the watcher never fires. For a field
+  with `secure: true` that meant a password crossing screens. Focus is dropped
+  on park too, so a field holding the keyboard does not arrive focused on the
+  next screen.
+
   Also: a parked sheet is dismissed and re-presented on return, rather than
   staying visible and interactive over the incoming screen — `.sheet` presents
   on the window, so the slot's `allowsHitTesting(false)` never reached it, and a
@@ -51,11 +60,15 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
   stack — a parked tab can be switched back to — and was released, throwing away
   the retention it should have kept.
 
-  The router now tags a stack-replacing navigation, and native reads the
-  animation from the prefix and the release from the suffix, so neither is
-  inferred from the other. No wire schema or NIF arity change: the transition
-  already travels as an atom whose name native reads directly, and the public
-  `:push | :pop | :reset` vocabulary callers use is unchanged.
+  The flag now rides on the JSON root, and the transition atom keeps its closed
+  `:push | :pop | :reset | :none` vocabulary. A first attempt suffixed the atom
+  instead (`:reset_replace`) and that was wrong: `set_transition/1`'s atom name
+  reaches Android as a raw string, and `MainActivity.kt` matches those four
+  values with an exact `when`, so every `reset_to` would have silently lost its
+  animation — on newly generated apps too, and those files are app-owned and
+  never re-rendered, so no template fix would have reached existing ones. An
+  unknown root key is ignored by both platforms' parsers, so a host that does
+  not read it keeps its current behaviour.
 
 - **A reset cross-fades again** (MOB-147). The outgoing screen was released
   synchronously, removing it in the same turn, so it hard-cut rather than fading
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,46 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
 
 ## [Unreleased]
 
+### Fixed
+- **A parked screen no longer keeps live resources running, and returning to
+  one no longer breaks the frame registry** (MOB-147, MOB-145). MOB-129 keeps
+  the navigated-away screen mounted so popping back diffs rather than rebuilds.
+  Before it, "not active" and "not alive" were the same state, so several things
+  never had to ask which one they were in. A post-merge review found two of them
+  silently broken rather than merely wasteful.
+
+  **The frame registry refused every write from a screen you popped back to.**
+  The nav generation is still bumped on every navigation and stale writes are
+  still refused, but a parked tracker's `onAppear` never fires again — so a
+  returning screen kept a stamp two navigations old and was rejected for ever.
+  `Mob.Test.element_frames` and `tap_id` read empty for the screen actually on
+  display, silently, for exactly the screens the optimisation retains. Trackers
+  now re-seed when their slot becomes active. Verified on device: 231 frames
+  before a push, 231 again after popping back.
+
+  **Toggles and sliders inherited state across screens.** Neither had any
+  prop-to-state sync, relying on the navigation teardown to re-seed them, so a
+  screen could show the toggle states and slider positions of the screen two
+  navigations back. `Mob.Socket` already raises `ArgumentError` on
+  `transition: :none` citing this exact hazard; MOB-129 had made it the default
+  path. Both now re-seed when the BEAM's value changes.
+
+  Also: a parked sheet is dismissed and re-presented on return, rather than
+  staying visible and interactive over the incoming screen — `.sheet` presents
+  on the window, so the slot's `allowsHitTesting(false)` never reached it, and a
+  park is no longer reported to the BEAM as a user dismissal. A parked video
+  pauses (audio included) and only resumes if it was autoplaying; a parked GPU
+  view stops rendering, having previously run at 60 fps indefinitely behind the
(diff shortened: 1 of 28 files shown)


GenericJam:
## Added: the remaining review findings (S1, S2), plus what S4 can and cannot answer

**S1 — the release keyed on the animation name, and was wrong in both directions.**

Native needs two independent facts: which animation to play, and whether the outgoing screen is still reachable. Both travelled as one atom, so:

- `reset_to(..., transition: :push)` — documented, **is** a stack replacement — was read as a push and its unreachable screen retained for ever
- `switch_tab(..., transition: :reset)` — documented, **is not** a replacement, since a parked tab can be switched back to — was released, throwing away exactly the depth-1 retention MOB-129 exists for

The router now tags a replacing navigation (`:push_replace` / `:pop_replace` / `:reset_replace`) and native reads the animation from the prefix, the release from the suffix.

**Encoded as a suffixed atom rather than a new wire field, deliberately.** `set_transition/1` already carries an atom end to end and the NIF reads its name directly — so no arity change (which would break the twelve test stubs that define `set_root/1`), no JSON schema change, and no change to the public `:push | :pop | :reset` vocabulary. An older native build falls to its default case: no slide, no release, a cosmetic degrade rather than a break.

**S2 — a reset cross-fades again.** The outgoing slot was cleared synchronously, removing it in the same turn, so it hard-cut rather than fading — and with `.transition()` gone there was nothing to animate a removal. The release now runs on the animation's completion.

### Tests

4 new for the split, plus **the tab-switch counterpart the old suite never had**: it asserted which animation reached native and never that a tab switch is not a replacement, which is half of what S1 got wrong. Four existing tests pinned the old contract and now pin the new one.

Mutations run — release keyed back on the animation name, offsets reading the raw atom, the router no longer tagging resets, release moved before the animation, the non-animated path never releasing — **all caught**. One did not fire: tagging `replacing(:none)`, because that clause is unreachable — a tab switch never goes through the reset path, which is the fix working rather than a test gap.

### S4 — partly answered, and I am not going to fake the rest

Steady state with **both slots warm**: 67.6 ms against 65.7 ms for one, about **3%**, which suggests the equality check is doing its job.

Rotation itself stays **unmeasured**. A container geometry change produces no `set_root`, so the frame instrumentation cannot see it at all — measuring it needs instrumentation that does not exist yet. Left open on MOB-147.

### Re-verified on device after the rework

| step | screen | `element_frames` |
|---|---|---|
| dense | DenseBenchScreen | 231 |
| push, then pop back | DenseBenchScreen | **231** |
| reset | HomeScreen | 1 |

1484 tests pass.

GenericJam:
## Review returned DO-NOT-MERGE. Three blocking findings, two of which I introduced. All fixed.

### B-1 — the `_replace` suffix broke Android

`set_transition/1`'s atom name reaches Android as a raw string, and `MainActivity.kt` matches `"push"/"pop"/"reset"` with an exact `when` — so `"reset_replace"` fell to `else` and **every `reset_to` silently lost its animation.** Newly generated apps too: the template was never updated, and those files are app-owned and never re-rendered, so a template fix wouldn't reach existing apps either.

My claim that *"an older native build falls to its default case — a cosmetic degrade"* was wrong twice over: it's **current** builds, and there's no upgrade path.

**The flag now rides on the JSON root** and the transition keeps its closed vocabulary. An unknown root key is ignored by both platforms' parsers. This is exactly what the suffix was avoiding, and it was the wrong trade — I optimised for not changing a NIF arity and broke a cross-platform contract instead.

### B-2 — the deferred release could blank the screen

`release` captured `outgoing` and checked only `releasing`. A reset followed inside 0.25 s by a push makes that captured slot the **active** one, so the completion cleared the slot the user was looking at — and nothing recovered until that screen repainted, which a mount-once screen never does. Guarded on `activeSlot != outgoing`, correct whether or not a superseded completion fires.

### B-3 — one call site missed, re-breaking the cross-fade the last commit claimed to fix

`slotOpacity[incoming]` still read the raw transition while everything else read the derived animation, so a reset cross-faded **or hard-cut depending on slot parity**. That four call sites needed coordinated edits and one was missed with no compiler or test signal is the strongest argument against the encoding, independent of Android.

### The previous B2 fix was incomplete in the common case

`onChange(of:)` fires on a **value** change. Slots alternate, so a screen reuses identities from two navigations back — and if both carry the same value (`false`, or `""`) it never fires. Toggle, slider and text field now re-seed on activation. `secure: true` renders a SecureField, so that was **a password crossing screens**. Focus is dropped on park.

### S-1 — the video resumed on every render, not on return

`updateUIViewController` runs on every update of the active screen and a fresh node graph arrives on every BEAM render, so a user's pause was undone within one render on any screen with a timer. Gated on the park-to-active transition.

### S-5 settled by experiment, not argument

Three park-direction fixes depended on whether `.equatable()` blocks environment propagation into the parked slot. Probe on a real GPU node:

```
[MobS5] gpu updateUIView isActive=1   ← on the screen
[MobS5] gpu updateUIView isActive=0   ← after parking
```

It propagates. The fixes are live, not dead code.

### Also

The comment stripper scans the file as one binary now, carrying string and block-comment state across lines — per-line scanning broke on Swift's `"""` literals (MobGpuView embeds MSL that way) and never recognised `/* */`, so commented-out code satisfied assertions.

**And an unrelated CI flake this surfaced:** `Mob.StateTest` failed in setup on a `Process.whereis` → `GenServer.stop` race. Pre-existing, and **the same shape was in 11 places across 10 files**, so it lands on whoever pushes next. Replaced with a shared helper.

12 mutations across the review fixes, all caught. Device-verified: 231 frames before a push and 231 after popping back, and reset-then-push 80 ms apart — inside the reset animation, the interleave that blanked the screen — lands correctly.

1489 tests pass.

GenericJam:
Second adversarial review is in and acted on. No blockers were found; the substantive findings were two state leaks and a set of test gaps that mutation testing confirmed were real.

**Tagging widened.** Only `reset_to` set `replaces_stack`, though `pop`, `pop_to`, `pop_to_root` and both failed-restart recovery branches also stop the outgoing screen. All five are tagged now. Tab switches deliberately are not — a parked tab can be switched back to, so its tree is the one case where retention actually pays.

Measured on both sides rather than argued:

| | tagged | untagged |
|---|---|---|
| the pop itself (8 pops, dense screen) | **39.4 ms** | 42.8 ms |
| the follow-on `push A→B; pop; push A→C` | **82.1 ms** | 84.4 ms |

The follow-on is the side that pays for the release, and it comes out as noise. The reason it is noise is that the retention being given up was not buying anything: C is a different screen from B, so what it replaces is a cross-shape diff, which this epic already measured as no cheaper than a fresh build. An earlier 18.5 ms reading for the tagged case is discarded in the decision record — different run conditions, did not reproduce.

**Two real leaks, both created by slot reuse:**

- A video player survived into a different screen. Slot reuse preserves view identity across screens by design, `makeUIViewController` does not run again, and `updateUIViewController` never reassigned `vc.player` — so screen C showed screen A's video, and `wasPlaying` from A's park started it playing with audio. The coordinator now tracks `src` and rebuilds on change.
- The `seeding` latch was order-dependent. Two programmatic writes straddling one SwiftUI pass left it armed for the wrong write and reported a change the user never made. Replaced with a stateless comparison against the BEAM's own value, so a re-seed is silent by construction.

**Test gaps, each confirmed by a surviving mutation before the fix:** swapping `navToken += 1` and `let token = navToken` made the entire release inert with the suite green; three of the four tagged sites had no coverage; the video test asserted a rename, so relaxing its condition to `} else {` restored exactly the semantics the change removed; `replaced_stack?` in the tab test read only the last painted frame, so it returned false when nothing was painted and the refutation passed vacuously; and the `:none` test asserted that `Mob.Socket` raises, which says nothing about the router path it was named for.

The comment stripper moved to `test/support/native_source.ex` with its own tests. Three holes closed, each of which silently corrupted the native assertions built on it: nesting is now Swift-only, so a `.m` comment mentioning a glob no longer swallows the rest of the file; raw strings are recognised; and a scan that does not end in code raises instead of returning output that reports comments as code.

Also narrowed `Mob.Sender.render/5,6` back to `atom()` — the tuple is substituted inside `handle_cast/2`, after the public function has been called, so it never crosses that boundary, and widening only that spec left `Mob.Screen.Server`'s matching specs inconsistent.

**Device verification.** iOS: cross-fade overlaps and settles clean, frame registration survives a pop-back (231 frames), the reset/push/pop interleave that defeated the old guard, no field bleed. Android: run end to end on a physical moto g from a freshly generated app — push, pop, reset and reset-with-push all land correctly and the cross-fade animates, which is the regression class the first `_replace` design failed. That gap is now closed.

1517 tests pass; credo and format clean. Filed MOB-148 separately for something I hit on the way: a freshly generated app cannot build for Android at all until you run `mix mob.icon` by hand.

Holding the merge for your call.

GenericJam:
Ran a full mutation sweep over every production change on the branch, not just the last round's. **Five survivors**, each a one-line change that made something inert with the suite green:

- `.onAppear` could stop calling `record(...)`. A view identity that disappears and returns — a lazy row scrolled off and back — then re-stamps its generation and registers nothing, so its id is missing from `element_frames` until an unrelated geometry change writes it. Same shape as B1, reached a different way.
- The activation re-seed could stop calling `record(...)` too: the regex stopped at the re-stamp, so half the fix was unpinned. Order matters as well, since a write made before the stamp is refreshed is refused.
- `record/2`'s lazy stamp could be made unconditional, which defeats the generation gate outright — an outgoing screen's mid-animation `.move` writes would re-stamp themselves current and be accepted, which is exactly what the gate refuses. The test that claims to keep the gate didn't cover it.
- The sheet's park branch could be deleted whole. The tests asserted only that `dismissedByPark` appeared somewhere, so a parked sheet staying presented over the incoming screen passed. Both directions are pinned now.
- `reset_to(scope: :all)` could stop being tagged — it discards every parked tab stack as well as the current one, so it replaces strictly more than the default, and had no coverage at all.

**Two flakes, and one of them mattered.** My own `:none` test asserted an exact frame list where an async paint lands one or two frames depending on timing; replaced with a bounded wait plus a property assertion. Chasing it turned up a real flake in `component_server_test.exs` (untouched by this branch): three sites did `Process.exit` then `Process.sleep(10)` and asserted. It fails roughly once in twenty full-suite runs and isn't seed-dependent. Converted to `Process.monitor` + `assert_receive {:DOWN, ...}`, the idiom the rest of that file already uses and a stronger guarantee since `terminate/2` completes before the process dies.

Worth flagging why that one is more than tidiness: a flaky suite silently corrupts mutation results. It mislabelled one of my own verdicts as "caught" before I noticed my baseline count was off by one — so the sweep was re-run comparing against an exact captured baseline rather than a hardcoded string.

Final state: **1521 tests**, all eleven high-value mutations caught against an exact baseline, credo and format clean, CI green. Six consecutive full-suite runs at different seeds with no failures.

Still holding the merge for your call.

PR #158, opened 2026-09-12
MOB-181: honor fixed_width / fixed_height on iOS Column and Row

GenericJam:
## Summary

- iOS Column and Row previously ignored `fixed_width` / `fixed_height` entirely (Android was already correct via `nodeModifier`); this applies the same fix pattern that mob #104 used for Box.
- Suppress the outer `.frame(maxWidth/maxHeight: .infinity)` on whichever axis a fixed dim pins so SwiftUI's chained-frame semantics don't hide the inner `.frame(width:height:)`.
- Fix the third place iOS emits a max-frame around a container: `MobLayoutWeight`. Without this, `fixed_width` on a `layout_weight`-flexed row would still be silently overridden, making the fix partial for the dominant Compose analogue.
- Documented precedence (fixed dims beat fill on iOS) in `decisions/2026-09-11-column-row-fixed-dims-precedence.md`.

## Test plan

- [x] `test/mob/native_column_row_layout_test.exs` — source-contract regression asserting each of the three frame blocks (Column, Row, `MobLayoutWeight`). Anchored to the surrounding case opener so a refactor that moves the block can't pass on the same tokens elsewhere in the file.
- [x] Verified the tests fail on partial revert: reverting only the swift changes fails all three; reverting only `MobLayoutWeight` fails test 3.
- [x] `mix test` (1788 passed, 38 excluded) — full suite green on the worktree.
- [x] `mix credo --strict` — no new issues.
- [x] Adversarial pre-commit review by a fresh subagent applied should-fix findings.
- [ ] **Device verification pending** — I can't drive a physical device from here; iOS-sim device-verify (`Mob.Test.frame/2` on Column/Row/weighted-row screens with fixed dims set) is needed before merge, same as #104's post-merge validation. Follow-on note in Linear MOB-181.

Linear: MOB-181
GH: closes mob#110

https://claude.ai/code/session_01Ezeu2FEZfRajeoL4ywEUDJ

### DIFF
--- a/decisions/2026-09-11-column-row-fixed-dims-precedence.md
+++ b/decisions/2026-09-11-column-row-fixed-dims-precedence.md
@@ -1,0 +1,69 @@
+# iOS: fixed_width / fixed_height beat fill_* on Column and Row
+
+- Date: 2026-09-11
+- Status: accepted
+- Linear: MOB-181
+- GitHub: mob#110
+
+## Context
+
+`MobNodeView` on iOS previously ignored `fixed_width` and `fixed_height` on
+Column and Row entirely. Column always applied `.frame(maxWidth: .infinity,
+maxHeight: fillHeight ? .infinity : nil)` and Row only applied a
+`fill_width`-driven `.frame(maxWidth: .infinity)`. Android was already
+correct via `nodeModifier`.
+
+The Box case had the same class of bug and was fixed in #104 (iOS honor
+fixed-height boxes). Column and Row are the remaining two containers with
+the same shape. The issue thread (mob#110) called out a precedence question
+that #104 dodged because Box has no `fill_*` props: when a caller sets
+`fixed_width: 100, fill_width: true` on a row, which wins? iOS and Android
+have to decide the same way or a screen will render differently.
+
+## Decision
+
+**Fixed dims beat fill on iOS.** When both `fixed_width` and `fill_width` are
+set on a Row, `fixed_width` wins. Same rule for `fixed_height` vs
+`fill_height` on Column, and for `fixed_*` vs `layout_weight` on the flexing
+axis (see `MobLayoutWeight` below). Rationale:
+
+1. It's internally consistent with the Box #104 fix: fixed always wins on
+   iOS, no exceptions.
+2. SwiftUI applies chained `.frame` modifiers outside-in — an outer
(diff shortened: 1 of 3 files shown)


GenericJam:
## Device-verified on physical iPhone SE (3rd gen), iOS 26.5.2

Ran the change on a physical iPhone SE 3rd gen (`00008110-001E1C3A34F8401E`) via `~/code/mob_plugin_demo` with its mob dep pointed at this branch. Screenshot attached to the session summary. Frames pulled via `Mob.Test.frame/2`:

| Case | Setup | Expected | Got | Verdict |
|---|---|---|---|---|
| A | `<Column width=200 height=120>` | 200 × 120 | 200 × 120 | ✅ PASS |
| B | `<Column height=80>` (fill_width default) | ? × 80 | 327 × 80 | ✅ PASS |
| C | `<Column width=100>` (content height) | 100 × ? | 100 × 33 | ✅ PASS |
| D | `<Row width=180 height=48>` | 180 × 48 | 180 × 48 | ✅ PASS |
| E | `<Row fill_width={true} width=150 height=40>` — precedence | 150 × 40 (fixed wins) | **150** × 40 | ✅ PASS |
| F1 | Column w/ `layout_weight=1 width=90` inside `<Row height=60>` | width=90 (fixed beats weight) | width=**90** | ✅ PASS |
| G | Baseline `<Column>` (no fixed dims) | fill width, content height | 327 × 33 | ✅ PASS (no regression) |

**E** confirms the precedence decision (fixed beats fill on iOS). **F1** confirms the `MobLayoutWeight` half of the fix — without it, `layout_weight=1` would have expanded the width to fill the parent, hiding the fixed_width=90.

**Not tested by this change (pre-existing behavior)**: F1's cross-axis height was 33 (content-hug), not 60 (parent height). `MobLayoutWeight` never applied a max-frame on the cross axis, so this is unrelated to MOB-181 — a weighted-child in a fixed-height row hugs on the cross axis unless it opts in with `fill_height`. Consistent with iOS `HStack(alignment: .center)` semantics.

Prop-name reminder for anyone repro'ing: the sigil attribute is `width={200} height={120}` (same as `<Image>`), not `fixed_width`/`fixed_height`. The `fixed_` naming is the internal Swift field name (`node.fixedWidth`). Documented for the future in this comment; if that's confusing enough to be worth a doc surface, happy to file separately.

PR #164, opened 2026-09-12
MOB-190: :anchored node type on iOS — floating panel at the root, Android arithmetic

GenericJam:
## What

A new `:anchored` node type on iOS, the half of Mishka Chelekom parity that lives in mob (the Android half is the mob_new bridge template, MOB-189).

- Children `[anchor, panel]`: the anchor renders in flow and publishes its bounds through an anchor preference; `MobRootView`'s root ZStack draws the panel above every Box and Scroll (`.overlayPreferenceValue`), so nothing on the way up can clip it. Not `.overlay` (clipped by ancestors) and not `.popover(isPresented:)` (adapts to a sheet on iPhone, self-dismisses). Rationale in `decisions/2026-09-12-ios-anchored-is-a-root-overlay.md`.
- Placement is `MobAnchoredPosition.origin/5`, a line-for-line port of the Android bridge's `MobAnchoredPositionProvider` (the web's `positionPopup()`): RTL mirroring, main-axis flip only when both conditions hold, raw nudge, clamp to the window (edge padding plus safe area) while the anchor is on screen. Props: `side`, `align`, `side_offset`, `align_offset`, `panel_offset_x/y`, `flip`, `clamp`, `edge_padding`, `panel_max_width/height`.
- `on_tap` on the node is the outside-tap dismiss request; the panel never closes itself, so its state cannot disagree with the screen's.
- `mob_nif.m` maps the type string and parses the props. Before this the type fell to the zero-initialised `MobNodeTypeColumn`, so every Mishka popover, tooltip, menu, select and combobox stacked inline.
- New file `ios/MobAnchored.swift` (the generated `build.zig` globs `ios/*.swift`; `MobRootView.swift` was already past swiftlint's file-length limit). `Anchored` on `priv/tags/ios.txt`; `guides/components.md` section; `test/mob/anchored_test.exs`.

## Verification

- `mix test` green (anchored + sigil + screen_case), format, credo, clang-format, swiftlint clean.
- iPhone 11 Pro Max simulator, Mishka showcase app built from this branch (`mix mob.deploy --native --ios`): the Popover page's panel floats above its trigger with the beak pointing at it; a tap outside flips `pop_details` to `false` through the screen.
- Adversarial review in progress; findings will be applied before merge.

## Not in this PR

- Android: `MobAnchored` lives in the mob_new bridge template (MOB-189). Until it lands, `Anchored` is iOS-only on the whitelist, which is the truth.
- A panel inside a presented `:sheet` is not seen by the root host (separate window); no Mishka component does that today.

Linear: MOB-190 (parent MOB-188)

https://claude.ai/code/session_017skG2QLaK9jFJBbJpW9VS8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,18 @@ Full module documentation: [hexdocs.pm/mob](https://hexdocs.pm/mob).
   snake_case atoms), read from the calling app's config at its compile time.
   Companion to mob_new #68, which lists the vendored Mishka Chelekom
   catalog's tags in a generated app's `config.exs`.
+- **`:anchored` node type on iOS** (MOB-190). Children `[anchor, panel]`: the
+  anchor renders in flow, the panel floats over the page at the root ZStack
+  (anchor preference + root overlay, so no `corner_radius` Box or Scroll on
+  the way up can clip it), placed by a verbatim port of the Android bridge's
+  `MobAnchoredPositionProvider` — `side` / `align` / `side_offset` /
+  `align_offset` / `panel_offset_*` / `flip` / `clamp` / `edge_padding` /
+  `panel_max_*`; `on_tap` on the node is the outside-tap dismiss request.
+  Before this `mob_nif.m` had no mapping for the type, so it fell to the
+  zero-initialised `MobNodeTypeColumn` and every Mishka Chelekom popover,
+  tooltip, menu, select and combobox stacked inline instead of floating.
+  `Anchored` joins `priv/tags/ios.txt`; the Android half lives in the
+  mob_new bridge template (MOB-189). Documented in `guides/components.md`.
 
 ### Fixed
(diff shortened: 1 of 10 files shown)


GenericJam:
Adversarial review applied (commit 7651ac0): parked screens publish no entries (`mobScreenIsActive`), the anchor keeps one structural position across open/close, the node's own frame/fill/padding/background/gestures apply as on Android, all scrims layer below all panels, decision record corrected (status-bar band divergence; the sigil does not report iOS-only tags — MOB-194).

Re-verified on the iPhone 11 Pro Max simulator with the Mishka showcase rebuilt from this commit:
- Popover page: panel floats above its trigger; outside tap flips `pop_details` to `false`.
- Parked-screen case from the review: opened the popover, popped back to Home via `Mob.Test.tap(:back)`, then tapped a Home button through the UI (`idb` tap, not the BEAM). The tap reached Home and pushed `TextScreen`, so no dismiss scrim survives the outgoing screen.

